PR #6, opened 2016-12-19
Rewrite for testability, error handling,  separation of concerns

nastevens:
First off, thank you for writing this library. The API is exactly at the level it should be for 99% of the interactions that people have with S3, as opposed to the extremely detailed APIs of something like rusoto or aws_sdk_rust.

What I found when working with the library, though, is that many potential sources of errors were handled with `panic!`. This was unworkable for my needs since I would be running this from a server. I started by rewriting chunks of the code to use the Rust `Result` type, and started to notice more and more places where I could "rustify" code. In the end I reworked quite a bit, and ended up with something that I think will be easier to use, maintain, and extend.

I'm opening this PR mainly as an option for you - if you like the changes I've made, great! I'd love to have this code live on in rust-s3. If not, no problem! I'll keep this code local to my own branch, and will absolutely make sure that you continue to get credit and proper license handling under MIT. I may opt to change the name to avoid confusion as well.

Here's a hitlist of changes:

  * All authentication is split out to a separate module, and could potentially be replaced by another library like warheadhateus if desired
  * S3 functions are now calls on the Bucket type
  * Stringly type region has been converted to an enum with appropriate conversion functions.
  * Unit tests added across all authentication functions.
  * Operational example added - simply requires S3 credentials and a bucket name to perform a CRUD test.
  * All functions and methods properly handle or return errors - no more panics or unwraps
  * openssl is replaced with rust-crypto to eliminate system library dependency
  * Docs for all public functions added, and `warn(missing_docs)` added at module level
  * Credentials are their own type and accept a Token for use with temporary credentials.
  * Requests use the bucket name in the URL path, not as a subdomain. This eliminates TLS issues for non-matching domain names.
  * List command now supports delimiters

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "rust-s3"
-version = "0.3.1"
-authors = ["Drazen Urch"]
+version = "0.4.0"
+authors = ["Drazen Urch", "Nick Stevens"]
 description = "Tiny Rust library for working with Amazon S3."
 repository = "https://github.com/durch/rust-s3"
 readme = "README.md"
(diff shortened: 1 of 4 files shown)


durch:
Hi @nastevens,

thank you for the very detailed PR, I'm glad this lib getting picked up :)

You are completely right about the panics, just one of many issues that needed addressing, I'll be sure to go through and merge this ASAP. I've been following your fork and saw some really nice code :)

In the meantime it would be great if you'd join us on [gitter](https://gitter.im/durch/rust-s3), we might actually have a community forming.

PR #11, opened 2017-03-04
Implement custom endpoints

sp0:
This pull requests adds the ability to use a provided endpoint, instead of using the amazon region endpoints. I personally needed this for local testing (using s3rver), rather than using the AWS servers.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -55,6 +55,7 @@ pub struct Bucket {
     credentials: Credentials,
     extra_headers: Headers,
     extra_query: Query,
+    endpoint: Option<String>,
 }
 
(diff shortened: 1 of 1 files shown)


durch:
Thnx for the PR, looks OK, the only problem I'm having is that it breaks compatibility, I've been planning to do a few more breaking changes, so I'd like to wait a bit more, and push all the changes together :)

sp0:
Hey looking closer into the details of this, I realised this  doesn't output the correct host in the http headers. So this is probably not the correct way to implement custom endpoints.

durch:
Its still an interesting feature, I'll keep something like this in mind roadmap-wise :)

PR #12, opened 2017-06-07
crypto: use RustCrypto utils over deprecated rust-crypto

posborne:
This change started due to my unfortunate exposure to a bug impacting
the ability for my project to target aarch64:
https://github.com/DaGenix/rust-crypto/issues/383.

In that issue (and in others) it is clear that rust-crypto is now
deprecated in favor of the RustCrypto crates (here we need sha2/hmac).
This change drops in this replacement for rust-crypto.

CC @nastevens


nastevens:
Pretty straight forward - LGTM

posborne:
@durch I believe has another larger change currently gated on this one that will be opened once this is merged if you can take a look.  That second change will remove the curl dependency which was giving us fits for some of our work on some embedded platforms.

durch:
@posborne thanks for this :), merged, and will be published to crates.io in a few minutes

durch:
@posborne, @nastevens published

PR #13, opened 2017-06-19
Large file sizes and missing fields when listing bucket contents

BMeu:
Hey!

I was trying to list the contents of a few buckets, but encountered a couple of problems:

- File sizes larger than `i32::MAX` cause Serde XML to complain about receiving a string instead of `i32` (in my case, I had a 6GB file that caused this), so I changed the type of `Owner.size` to `u64`.
- For some buckets, the fields `Owner`, `NextMarker`, `Delimiter`, `Marker`, `EncodingType`, and `CommonPrefixes` were missing in the API response (again, causing Serde XML to fail), so I made them optional.

These are of course breaking changes since they change public types, but for me they work ;-)

### DIFF
--- a/src/serde_types.rs
+++ b/src/serde_types.rs
@@ -28,7 +28,7 @@ pub struct Object {
     pub key: String,
     #[serde(rename = "Owner")]
     /// Bucket owner
-    pub owner: Owner,
+    pub owner: Option<Owner>,
     #[serde(rename = "Size")]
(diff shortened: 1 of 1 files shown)


deg4uss3r:
+1 for optional fields :)

durch:
@BMeu, sorry completely missed this PR, will merge.

PR #15, opened 2017-08-03
Fixed the bucket list API

golddranks:
The bucket list API is currently dysfunctional. AWS sends responses that fail to parse, as Serde expects the fields to exist in responses. Additionally, when the request fails, the developer doesn't get very good information why.

This PR fixes the bucket list API by making non-default fields optional. In addition, it makes the errors returned by AWS into first class errors in the library.

The other methods need to be updated with AWS errors too – this is just the first push, but at least it makes the bucket listing API work.

### DIFF
--- a/src/bucket.rs
+++ b/src/bucket.rs
@@ -157,7 +157,7 @@ impl Bucket {
         let request = Request::new(self, "/", command);
         let result = request.execute()?;
         let result_string = String::from_utf8_lossy(&result.0);
-        if 200 <= result.1 && result.1 < 300 {
+        if result.1 < 300 {
             let deserialized: ListBucketResult = serde_xml::deserialize(result_string.as_bytes())?;
             Ok((deserialized, result.1))
         } else {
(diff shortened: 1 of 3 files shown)


durch:
@golddranks thnx for this, glad it is gettig some love, I'll review and merge, you have some changes that overlapp with #13, I'll sort it out.

golddranks:
Thanks <3

src/bucket.rs
```diff
@@ -157,8 +157,17 @@ impl Bucket {
         let request = Request::new(self, "/", command);
         let result = request.execute()?;
         let result_string = String::from_utf8_lossy(&result.0);
-        let deserialized: ListBucketResult = serde_xml::deserialize(result_string.as_bytes())?;
-        Ok((deserialized, result.1))
+        if 200 <= result.1 && result.1 < 300 {
```
durch: Couldn't we just change this to `< 300`, or even `< 400`, not sure if there are legit redirects possible.

PR #16, opened 2017-08-19
Upgrade dependencies

beanieboi:
Hey,

I'm starting to learn rust and I use rust-s3 in one of my first projects. I saw that the dependencies are pretty old and I got a lot of duplicate dependencies on my project. so I gave it a shot and looked how much work it is to upgrade the dependencies to the current latest versions. it wasn't that hard.

let me know what you think.


durch:
Hey,

Amazing, thnx for taking it on, you never know how these upgrades will go :)

PR #20, opened 2017-09-13
Moved error handling to the Request object, so every method now checks for errors.

golddranks:
Last time I implemented an error handling for one function. Now that I'm using many other functions too, I realised, upon encountering an error, that I completely forgot about the others!

You can always see from the HTTP status returned by AWS that some error has happened, if its 400 or over. So it makes sense to implement a general error handling. Also, the AwsError type contains the fields that, according to the API, are always populated. This is enough to get a gist of of the error.

At the moment the methods in this crate happily return with the "no error" path, even if error has happened. This is very surprising from the user perspective, since the methods return a `S3Result`. This PR fixes this by checking the errors at the `request.rs` file, upon finishing a request.

### DIFF
--- a/src/bucket.rs
+++ b/src/bucket.rs
@@ -1,14 +1,14 @@
 use std::collections::HashMap;
 use std::mem;
+
 use serde_xml;
 
 use credentials::Credentials;
 use command::Command;
 use region::Region;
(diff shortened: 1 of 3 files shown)


golddranks:
Oops, sorry, this was against master.

PR #31, opened 2018-07-19
Allowing custom region to optionally provide URL scheme

bwtril-justin:
The hard-coded URL scheme ([https://](https://github.com/durch/rust-s3/blob/eb247f8bf5fc922d1506be0b6e8956b3e6a8c29f/src/request.rs#L50)) in use today precludes or hinders the use of some S3-compatible APIs, including hosted solutions such as [LocalStack](https://localstack.cloud/) or [Minio](https://www.minio.io/).  The change introduced by this PR checks for the presence of a scheme (determined by the presence of `://`) within `region::Region::Custom(String)`.  Should one be located, the string will be used as-is to compose the scheme and host portion of the S3 endpoint.  Otherwise, the original behavior (hard-coded `https://` plus host) is applied.

Tests demonstrating original and new behavior are provided.

### DIFF
--- a/src/request.rs
+++ b/src/request.rs
@@ -47,8 +47,10 @@ impl<'a> Request<'a> {
     }
 
     fn url(&self) -> Url {
-        let mut url_str = String::from("https://");
-        url_str.push_str(self.bucket.host());
+        let mut url_str = match self.bucket.host().find("://") {
+            Some(_) => String::from(self.bucket.host()),
+            None => format!("https://{}", self.bucket.host())
+        };
         url_str.push_str("/");
         url_str.push_str(self.bucket.name());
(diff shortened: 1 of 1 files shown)


bwtril-justin:
It was discovered that the URL host was including the scheme.  The modifications to the tests reflect this:

```
thread 'request::tests::url_uses_scheme_from_custom_region_if_defined' panicked at 'assertion failed: `(left == right)`
  left: `"http://custom-region"`,
 right: `"custom-region"`', src/request.rs:268:9
```

durch:
Thnx @bwtril-justin, sorry for the wait, missed the notification, new version published to crates.io

PR #34, opened 2018-09-24
Convert to using reqwest library

nastevens:
When this library was originally created there wasn't a good pure-Rust solution for synchronous HTTP requests. That library (reqwest) now exists, so convert existing Curl-based code to use it.

This primarily reduces the number of external system/C-language dependencies for the library, but the reqwest library also has much nicer ergonomics than curl.

@durch

---
This change is [Reviewable](https://reviewable.io/reviews/durch/rust-s3/34)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,10 +15,10 @@ path = "src/lib.rs"
 
 [dependencies]
 chrono = "^0.4.0"
-curl = "^0.4.8"
 error-chain = "^0.10.0"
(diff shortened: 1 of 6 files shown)


jmagnuson:
This seems like a good step forward, unless it was decided that [rusoto](https://github.com/rusoto/rusoto) or something should get used under the hood instead. And I had issues using `no-verify-ssl` with the the curl implementation, but this one works fine.

PR #38, opened 2019-03-09
GetTags command

jgoday:
GetTags method to get the current tags from a file [RESTObjectGETtagging](https://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectGETtagging.html)
The return type is Vec<String, String> (a vector of tagname->tagvalue tuples) and the operation result code `S3Result<(Vec<(String, String)>, u32)>`.

XML 'parsing' (for this simple purpose) was done using a simple regular expression `(r"<Key>([\s|\S]+)</Key><Value>([\s|\S]+)</Value>")`.

Added [**regex**](https://docs.rs/regex/1.1.2/regex/) external crate .

---
This change is [Reviewable](https://reviewable.io/reviews/durch/rust-s3/38)


durch:
Published version `0.12.1`

PR #48, opened 2019-09-03
Finalize 0.18.0-beta

### LINKED ISSUE #47: Async version of `list` command?
I'm trying to use `rust-s3` in a method returning a Future to asynchronously get data from an S3 bucket and am currently using version `0.16.0-alpha2`. Using `get_object_async` is already working fine for me, however, I would now like to list the contents of the Bucket to download further objects in the same async function. However, if I call `mybucket.list`, it seems that the list method is not currently compatible with asynchronous operation:

```
executor failed to spawn task: tokio::spawn failed (is a tokio runtime running this future?)
```

Would it be possible to add a `list_async` function similar to the `get_object_async` method?

Many thanks for providing this very useful library!

durch:
Closes #47. 

Adds async version of the `list` bucket method (`list_all_async`), renames `list` to `list_all`. Also adds `list_page` and `list_page_async` `Bucket` methods. Simplifies returned futures so that only a top level future remains, making usage much more ergonomic.

---
This change is [Reviewable](https://reviewable.io/reviews/durch/rust-s3/48)

### DIFF
--- a/src/bucket.rs
+++ b/src/bucket.rs
@@ -7,28 +7,13 @@ use std::io::Write;
 use credentials::Credentials;
 use command::Command;
 use region::Region;
-use request::{Request, Headers, Query, S3Error};
+use request::{Request, Headers, Query};
 use serde_types::{ListBucketResult, BucketLocationResult};
 use serde_types::Tagging;
 use futures::Future;
-use futures::future::{loop_fn, Loop, ok, result, LoopFn};
-use snafu::{ResultExt, Snafu};
+use futures::future::{loop_fn, Loop};
+use error::{S3Error, S3Result};
 
-#[derive(Debug, Snafu)]
-pub enum Error {
-    XmlError {
-        source: serde_xml::Error
-    },
-    RequestError {
-        source: ::request::Error
-    },
-    ParseError {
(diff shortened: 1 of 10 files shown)


src/bucket.rs
```diff
@@ -443,64 +443,65 @@ impl Bucket {
...
+    /// let region = "us-east-1".parse().unwrap();
+    /// let credentials = Credentials::default();
+    /// let bucket = Bucket::new(bucket_name, region, credentials).unwrap();
+    ///
+    /// let results = bucket.list_all_async("/", Some("/")).and_then(|list| {
+    ///     println!("{:?}", list);
+    ///     Ok(())
+    /// });
+    /// ```
+    pub fn list_all_async<'a>(&'a self, prefix: &'a str, delimiter: Option<&'a str>) -> impl Future<Item=Vec<ListBucketResult>> + 'a {
```
sseemayer: Does anything speak against making this `impl Future<_> + 'a + Send` ?

sseemayer: Hang on, I didn't need `Send` when getting objects so I don't think I need it here either. Will experiment some more.

PR #68, opened 2019-12-30
Fix to avoid unintentional fails

### LINKED ISSUE #67: Crash with "Not an EC2 instance", even in EC2
Hi, I found it sometimes crashes with "Not an EC2 instance", even in EC2 instances.

According to https://github.com/scylladb/scylla/commit/2a4ba883c8e9c3351ff69064bb4d09954ff001c4#diff-23bb822fd3271daf086e185150f31517, it's because some EC2 instances don't have `/sys/hypervisor/uuid`, but have `/sys/class/dmi/id/board_vendor`.

kenkoooo:
Currently, `rust-s3` reads `/sys/hypervisor/uuid` to determine whether the instance is EC2 or not. But some EC2 instances don't have `/sys/hypervisor/uuid`, but have `/sys/class/dmi/id/board_vendor`. This change will enable `rust-s3` to run also in the latter type EC2 instances.

Reference:
https://github.com/scylladb/scylla/commit/2a4ba883c8e9c3351ff69064bb4d09954ff001c4#diff-23bb822fd3271daf086e185150f31517

Closes #67

---
This change is [Reviewable](https://reviewable.io/reviews/durch/rust-s3/68)

### DIFF
--- a/src/credentials.rs
+++ b/src/credentials.rs
@@ -121,14 +121,8 @@ impl Credentials {
     }
 
     fn from_instance_metadata() -> S3Result<Credentials> {
-        let output = Command::new("sh")
-            .arg("-c")
-            .arg("head -c 3 /sys/hypervisor/uuid")
-            .output()
-            .expect("failed to execute process");
-        let is_ec2 = std::str::from_utf8(output.stdout.as_slice())? == "ec2";
-        if !is_ec2 {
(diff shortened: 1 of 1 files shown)


durch:
@kenkoooo thank you for the PR, it will be on crates.io in a few minutes, version 0.18.9

PR #82, opened 2020-04-28
update import of S3Error from root to new mod

dvente:
S3Error was still being imported from the root but it was moved to the error module, so I updated the import


durch:
@dvente Thank you :)

PR #86, opened 2020-05-01
Remove debug println!

npv3s:
I used this library and suddenly I saw an incomprehensible link from this file


durch:
@npv3s thank you :)

PR #89, opened 2020-05-10
caching reqwest client instance as advised by reqwest documentation

Lesiuk:
The reqwest documentation tells us that we should reuse client istance as it holds connection pool internally

https://docs.rs/reqwest/0.10.4/reqwest/struct.Client.html


durch:
@Lesiuk thank you for this, I'll review and merge promptly

PR #91, opened 2020-05-25
tokio_get_object_stream should not buffer whole file in memory

### LINKED ISSUE #90: tokio_get_object_stream is not streaming data but buffering everything in memory and writing all at once
It's going to be faster for large files download and also use significantly less memory.

Lesiuk:


durch:
@Lesiuk very nice! Thank you

durch:
Version `0.22.4` published to crates.io

PR #92, opened 2020-05-27
Fix continue logic of list_buckets

shonenada:
Hi, 

> NextContinuationToken is sent when isTruncated is true, which means there are more keys in the bucket that can be listed. 

According to [S3 docs](https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectsV2.html#AmazonS3-ListObjectsV2-response-NextContinuationToken), it should ensure `isTruncated` to be `true` before using `NextContinuationToken`, or it will cause forever loop (in my case).

@durch Please review.


durch:
@shonenada, thank you for the PR, published to crates.io, version 0.22.5

PR #93, opened 2020-06-17
handle S3 URLs that are encoded differently in percent-encoding and AWS

njaard:
the Url crate does not encode (for example) parentheses, but AWS
wants them for the purposes of signing. This change decodes
the Url and then re-encodes it with AWS's idea of canonical
percent-encoding.


durch:
@njaard thank you for the fix, I've just hit this issue last week! Funny how that works. Published in version `0.22.7`

PR #95, opened 2020-06-23
Fixing `PresignPut`

shonenada:
Hi @durch .

I'm using rust-s3 for my project, and I found that the following codes will panic my program with unreachable exception.

```
let url = bucket.presign_put("/test.file", 86400).unwrap();
```

After review the codes I found something not correct, but I'm not sure whether there should be `Command::PresignPut` arm in `fn presigned_canonical_request`, so plz review.

Besides, the `Host` header in `presigned_authorization` should be generated according to the `cfg!(feature = "path-style")` .

### DIFF
--- a/s3/src/request.rs
+++ b/s3/src/request.rs
@@ -64,8 +64,8 @@ impl<'a> Request<'a> {
 
     pub fn presigned(&self) -> Result<String> {
         let expiry = match self.command {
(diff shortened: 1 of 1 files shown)


durch:
@shonenada thank you for the PR, presign_put was a last minute addition there, will review and merge promptly!

durch:
Published as `0.22.9`

PR #99, opened 2020-07-10
Bucket path style

### LINKED ISSUE #96: Path-style configuration in runtime
Hi there ! 

Thanks for your work on this bucket :) Really like the simplicity of it compared to other existing solutions in Rust ! 
Actually enjoyed integrating S3 into an app. Was definitely not the case using rusoto... 

Anyway, I would like to suggest making the `path-style` cargo feature a configuration flag. The problem with using it as a cargo feature flag is one cannot use multiple backends in the same app. Which is my case at the moment. Perhaps, it would be possible to add an optional boolean value to the `Region` to specify if it is path based or subdomain-based. 

Will try to make something work locally and try to send a PR in !

durch:
Closes #96.

Heads up @iammichiel this PR addresses the issue of runtime `path-style` configuration

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -2,20 +2,22 @@ name: build
 
 on:
   push:
-    branches: [ master ]
+    branches: [master]
   pull_request:
-    branches: [ master ]
+    branches: [master]
 
 jobs:
   build:
-
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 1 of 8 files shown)


durch:
`v0.22.11` published to crates.io

iammichiel:
Awesome ! Didn't even have time write a dirty hack and this is already merged 👍

PR #107, opened 2020-07-29
Fix put_object_stream, add tests, closes #106

### LINKED ISSUE #106: put_object_stream_blocking uploads 0B file
**Describe the bug**
When using `put_object_stream_blocking`, I went to check the results of the file upload only to find 0B file of the correct name.

**To Reproduce**
Using a String or Path, open the file

```
// Open and upload file
let path = Path::new(&s.file);
if !path.exists() {
  error!("{} not found!", path.to_str().unwrap());
  std::process::exit(1);
}

let mut file = File::open(path).expect("Unable to open file!");

let status_code = bucket
.put_object_stream_blocking(&mut file, &server_filename)
.unwrap_or_else(|e| {
  error!("Unable to upload {}! Error: {}", &server_filename, e);
  std::process::exit(1);
});

println!("Status code: {}", status_code);
```
**Expected behavior**
The file is currently 180KB. I expect that to be reflected in what's uploaded to AWS.

**Environment**
 - Rust version: 1.45.0
 - lib version: 0.22.12

**Additional context**
Context: I'm uploading a firmware binary file.

durch:
This was an embarrassing bug :). `put_object_stream` was using `std::io::Read` to read a file, which resulted in `0B` files being uploaded. Fixed by making it generic over `futures::io::AsyncReadExt`.

This in turn affected the put_object_stream_blocking` as it was simply blocking on the async function.


jplatte:
Why would you make anything generic over `TraitExt` rather than `Trait` (usually the former is implemented for any `T: Trait`)?

durch:
@jplatte thanks for the nudge :). I've been away for about two weeks, will pick it up soon, and yeah, final state will be generic over the `Trait` as opposed to `TraitExt`

Issue #111, opened 2020-09-17
Tests assume some environment variables to be present


jplatte:
**Describe the bug**

When running `cargo test` for the s3 crate in an environment without the EU_AWS_ACCESS_KEY_ID & EU_AWS_SECRET_ACCESS_KEY environment variables, the tests fail.

**To Reproduce**

```sh
unset EU_AWS_ACCESS_KEY_ID
unset EU_AWS_SECRET_ACCESS_KEY
cd s3
cargo test
```

**Expected behavior**

Tests should succeed.

**Environment**
 - Rust version: `1.46`
 - lib version: git master

jplatte:
Actually, this seems even worse. Do these have to be valid credentials for the tests to pass?

durch:
@jplatte yeah, I did not have the bandwidth to mock the tests. Github actions associated with the repo pull in those secrets, if I recall correctly they get pulled into forked repo actions as well.

durch:
Actually judging by the output of that PR they don't, now that is extra crappy, I'll pull your branch over to test

durch:
@jplatte thank you for raising this, sorry for the crappy PR experiance, I'll have to figure something out here. I've merged your bumps, its published to crates.io as `0.24.0`, I'll keep this issue open.

jplatte:
If they were available to PR branches, they could be leaked through PRs that update the actions workflow.

Maybe you could add a `realworld-tests` cargo feature and add `#[cfg(not(feature = "realworld-tests"), ignore]` to these tests (+ update your CI config to pass `--features realworld-tests`).

PR #112, opened 2020-09-17
Bump dependencies

jplatte:
Tests don't all succeed locally because of #111, but the signing ones do so I think everything is correct.


jplatte:
Okay so the same tests that failed locally now fail here, with a different error message. Is this caused by the CI secrets not being available for PRs from people without the necessary permissions?

PR #119, opened 2020-09-23
Allow types other than string slices for local paths

jplatte:
Before this, local paths obtained through directory traversal, `tempfile::NamedFile::path` or anything like that would first have to be converted to strings, which does UTF8 validation and is just generally pointless boilerplate.

This is a breaking change since it can break type inference (for example if people were calling `.put_object_stream(my_local_path.as_ref(), my_s3_path)`).

### DIFF
--- a/s3/src/bucket.rs
+++ b/s3/src/bucket.rs
@@ -22,6 +22,7 @@ use tokio::io::AsyncWrite as TokioAsyncWrite;
 use reqwest::header::HeaderMap;
 
 use async_std::fs::File;
+use async_std::path::Path;
 
 /// # Example
 /// ```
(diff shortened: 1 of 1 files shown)


jplatte:
Once std has `AsyncRead` it would probably make sense to also have a variant of `put_object_stream` that takes any `AsyncRead`. You could do this now with `async-std`s `AsyncRead`, I guess it's not much different from this PR which also exposes an `async-std` type in the public API (though I can change it to `std::path::Path` if you want, which is convertible to `async_std::path::Path`, if you prefer).

jplatte:
Rebased. Ping @durch

durch:
@jgoday I have no strong opinions here, as long as we keep making the API more ergonomic. Having a `pub_object_stream` generic over `AsyncRead` would be nice for sure :)

jplatte:
Awesome. By the way, how about setting up a dev chat? (I personally prefer Matrix, Zulip would be cool too)

durch:
@jplatte published to crates.io as `0.26.0`

PR #133, opened 2020-11-02
Presigned PUT requests should not encode path

iammichiel:
Hi there, 

While trying to perform a presigned PUT upload, I noticed the following : 
- uploads to a path without a `/` work fine and uploads with a `/` in path fails. 

So I started digging around and it seems the path is url-encoded when performing a presign. Which, I think should not be the case... 
The presign request itself is working fine, the actual upload PUT request will fail with a 403. 

The suggested change actually corrects the problem for my storage object provider (locally on Minio and online at Scaleway).

### DIFF
--- a/s3/src/request.rs
+++ b/s3/src/request.rs
@@ -234,7 +234,7 @@ impl<'a> Request<'a> {
         };
         let url = Url::parse(&format!(
             "{}{}",
-            self.url(true),
+            self.url(false),
             &signing::authorization_query_params_no_sig(
                 &self.bucket.access_key().unwrap(),
                 &self.datetime,


durch:
@iammichiel  think the issue might be with the `path_style` url `minio` is using. Did you init your bucket with [Bucket::new_with_path_style()](https://docs.rs/rust-s3/0.26.3/s3/bucket/struct.Bucket.html#method.new_with_path_style)?

iammichiel:
Well, actually locally, I am using the path_style for local development but deployed online, I am using the subdomain method with a check : 

```rust
    let storage: Bucket = if storage_type_path {
        Bucket::new_with_path_style(&storage_bucket, region, credentials).unwrap()
    } else {
        Bucket::new(&storage_bucket, region, credentials).unwrap()
    };
```

The suggested fix is working both locally and online. I am setting up an AWS S3 env to check this behaviour on AWS (the source of truth!).

durch:
Happy to hear that :)

iammichiel:
Ok, so finally had some time to check this. So a clear behaviour has come up. 
When presigning a key using the S3 api, you can either url-encode the key or not. 

When **encoding**, currently `rust-s3` behaviour : 
- works on AWS 
- works on Minio 
- does not work on Scaleway

When not encoding the key : 
- works on AWS, 
- works on Minio, 
- works on Scaleway 

So, not sure what the correct behaviour is so far. I have opened a support ticket with Scaleway. Will get back as soon as I get feedback from them.

durch:
Thank you for looking into this!

iammichiel:
Okay, so they finally answered my ticket but said the problem was ACL related. However, without changing anything on my side, all of a sudden, things started to work :) 

Let's consider this might have been a bug on their side and close this ticket ! 🎉

Issue #134, opened 2020-11-02
Support for range in get object


guilload:
Hello,

I need support for range in get object for a project I'm working on. I'm happy to submit a PR. Do you want me to add a `range: Option<(u64, u64)>` parameter to the `GetObject` command or create a new `GetObjectRange` command? I guess I'd also have to create a new method on the `Bucket` struct called something like `get_object_range`. Shall I also implement the blocking API?

Let me know what you think.

durch:
@guilload that is a good question, I feel like a `GetObjectRange` command is more ergonomic, than another parameter for the `get_object` method family. If its all the same to you I'd go with a new command. I think the only change you'll need to introduce is a conditional header if its the  `GetObjectRange`.  Let me know what you think :)

Blocking API is currently leveraging `tokio` and is only doing a `Runtime::block_on`, I've got a rework of that in progress. There is no need to implement a separate blocking method as these will soon-ish get auto implemented.

durch:
Published to crates.io in `0.27.0-beta0`

PR #136, opened 2020-11-03
Implement get object range

guilload:
Tentative implementation for #134. Hopefully, I got the [spec](https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.35) right. Let me know what you think.


durch:
@guilload thank you, will pull this into `dev`, it'll likely go out in `0.27`. I'll tag the release here as well

PR #139, opened 2020-11-13
Fix `Credentials::from_instance_metadata`

guilload:
Additionally, make the method public and remove the need for cloning some strings.


durch:
Looks great, thank you!

PR #140, opened 2020-11-13
Improve `get_object_range` documentation

guilload:
The range is inclusive: `[start, end]` instead of `[start, end)`. I ran into an issue because of this, so I thought I'd make the documentation clearer.


durch:
Very nice, thank you!

PR #142, opened 2020-11-18
add: impl create_with_path_style

extraymond:
Addition to #99, so creating buckets with custom s3-conformance servers work.

Extra info:
Prior to the current beta, it's possible to create a bucket with path style by building bucket and using command directly, but it's not feasible in the latest upstream. IIRC, it's due to request not being public.


durch:
@extraymond Thanks for the PR, pushed out in `0.27.0-beta1`

Issue #145, opened 2020-12-01
the trait `Send` is not implemented for `Result<(Vec<u8>, u16), s3::S3Error>`


godofdream:
**Describe the bug**
dyn Future<Output = std::result::Result<(Vec<u8>, u16), s3::S3Error>>` cannot be sent between threads safely
```
error[E0277]: `dyn Future<Output = std::result::Result<(Vec<u8>, u16), s3::S3Error>>` cannot be sent between threads safely
   --> src/main.rs:39:9
    |
39  |         tokio::spawn(async move {
    |         ^^^^^^^^^^^^ `dyn Future<Output = std::result::Result<(Vec<u8>, u16), s3::S3Error>>` cannot be sent between threads safely
    | 
   ::: /home/xxx/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-0.3.5/src/task/spawn.rs:128:21
    |
128 |         T: Future + Send + 'static,
    |                     ---- required by this bound in `tokio::spawn`
    |
    = help: the trait `Send` is not implemented for `dyn Future<Output = std::result::Result<(Vec<u8>, u16), s3::S3Error>>`
```

**To Reproduce**
simply use tokio spawn and it crashes during build.

**Environment**
```
aws-creds = { version = "0.25", default-features = false, features=["rustls-tls"]}
aws-region = "0.22"
rust-s3 = {version = "0.27.0-beta1 ", default-features = false, features = ["with-tokio","tokio-rustls-tls"], as="s3"}
```

**Additional context**
I assume "simpl::err" doesn't add "Send" trait

durch:
@godofdream interesting issue,  merged you PR over in `simplerr` prematurely. The issue is that Error trait source method does not allow the Send + Sync boundary. Could you provide a code snippet that I could start on, it'll make experimantation easier :)

durch:
Tentatively merged to `dev`, published beta3 to crates.io

PR #146, opened 2020-12-02
switch to anyhow error

godofdream:
switched to anyhow error and replaced some match statements with function chaining.

using the anyhow error should fix https://github.com/durch/rust-s3/issues/145


godofdream:
build fails, because i bumped all 3 packages

durch:
@godofdream nice one! Thank you, I should be able to merge + publish this today

PR #154, opened 2020-12-25
Delete a dependency on async-std when tokio is selected

nightmared:
Hello, this tiny PR removes the dependency on async-std when he feature flag `with-tokio` is enabled (in the current state, this means that we are pulling the async-std runtime and dependency tree even when we only make use of the tokio runtime).

Happy holidays, and many thanks for this crate, by the way :)

### DIFF
--- a/s3/Cargo.toml
+++ b/s3/Cargo.toml
@@ -49,7 +49,7 @@ url = "2"
 block_on_proc = { version = "0.1", optional = true }
 
 [features]
-with-tokio = ["reqwest", "tokio", "futures", "async-std"]
+with-tokio = ["reqwest", "tokio", "futures", "tokio/fs"]
 with-async-std = ["async-std", "surf", "futures"]
 sync = ["attohttpc", "maybe-async/is_sync"]
 default = ["tokio-native-tls"]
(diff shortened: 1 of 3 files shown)


nightmared:
Hmm, I'm not quite sure the PRs failures are related to the commit:
```
---- bucket::test::test_bucket_create_delete_default_region stdout ----
thread 'bucket::test::test_bucket_create_delete_default_region' panicked at 'assertion failed: `(left == right)`
  left: `"<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n<Error><Code>AuthorizationHeaderMalformed</Code><Message>The authorization header is malformed; a non-empty Access Key (AKID) must be provided in the credential.</Message><RequestId>7SEYCNEG9G7K7MFT</RequestId><HostId>4OaxZZ2dEYVNrav9WwpYVmwsRtnSl8RGyNdu1oc6Mso9Um+6jvx1dogWp6A1xMb6VazArdP+YvA=</HostId></Error>"`,
 right: `""`', s3/src/lib.rs:1:1
 ```
 At the very least, "make ci" runs fine on my computer, and I obviously cannot "make ci-all" because I (rightfully) don't have the s3 credentials.

durch:
@nightmared great stuff, I've been meaning to do that for a while now. Apologies for the CI, I've not figured out a sensible way to test against actual services for external PRs.

durch:
Published as `0.27.0-beta8`

PR #155, opened 2020-12-25
add custom headers on put_object requests

nightmared:
Hello again.

I'm quite lazy, so I didn't want to upload objects with a presigned put (which allows for passing custom metadata). This means I had to be able to pass custom headers (in theory I only needed to add "x-amz-meta-*", but while we are at it, wa can as well let the user add any possible header, right ?). This is an attempt at solving this problem.
However, I do think that this PR is less than ideal because
1) a builder pattern would probably prevent adding too many similar functions, and be better than this hacky patch
2) we forward `custom_headers` to the s3 endpoint both for submitting parts & for initiating the upload, without distinguishing both (in the event of a multi-parts upload of course, otherwise there is no such dichotomy because there is only a single request). However adding two arguments would be tedious and there is no way to distinguish the headers targeting the parts from those targeting the `InitiateMultipartUpload` query. So I decided to send the same headers for both.

Anyway, do you think something like that would be valuable ?

### DIFF
--- a/s3/src/bucket.rs
+++ b/s3/src/bucket.rs
@@ -565,7 +565,7 @@ impl Bucket {
     /// // Blocking variant, generated with `blocking` feature in combination
     /// // with `tokio` or `async-std` features.
     /// #[cfg(feature = "blocking")]
-    /// let status_code = bucket.put_object_stream_blocking(&mut path, "/path")?;
+    /// let status_code = bucket.put_object_stream_blocking(&mut file, "/path")?;
     /// #
     /// # Ok(())
     /// # }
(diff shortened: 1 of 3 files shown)


durch:
@nightmared sorry about dragging over here, the dealing with some quakes in my part of the world, will get to it by EoW hopefully

nightmared:
@durch no worries, take all the time you want, I have no deadline whatsoever on my end (and I'm not paying you anyway, so I'm not entitled to an answer either, don't fret !).

nightmared:
Mmh there is quite a few conflicts, let me rebase that.

nightmared:
Considering this was closed, does this mean you expect me to reopen this PR against master, or that you don't want this feature?

PR #157, opened 2021-01-01
Publicize direct stream reader for s3 putobject

Protryon:
Necessary for using streams that are not files.


durch:
@Protryon thanks for the PR will publish in `0.27.0-beta9`

durch:
Looking at the PR, I'm thinking we should simply make put_object_stream generic over streams, as opposed to having an implicitly specialized function, will give it some thought. Should not be a big deal, tests might be the biggest pain to deal with...

durch:
Made `put_object_stream` generic over `AsyncRead|Read`, will be published to crates.io today

durch:
published in `0.27.0-beta9`

PR #160, opened 2021-01-06
upgrade tokio to 1.0

### LINKED ISSUE #158: Tokio = 1.0
Tokio's current version is 1.0. It would be useful that this crate will use that version as well

nightmared:
Fixes #158

### DIFF
--- a/s3/Cargo.toml
+++ b/s3/Cargo.toml
@@ -35,15 +35,15 @@ log = "0.4"
 maybe-async = { version = "0.2" }
 md5 = "0.7"
 percent-encoding = "2"
-reqwest = { version = "0.10", default-features = false, features = ["json", "stream"], optional = true }
+reqwest = { version = "0.11", default-features = false, features = ["json", "stream"], optional = true }
 serde = "1"
 serde_derive = "1"
 serde-xml-rs = "0.4"
 sha2 = "0.9"
 anyhow = "1.0"
 surf = { version = "2", optional = true, default-features = false, features = ["hyper-client"] }
-# Waiting on h2 compat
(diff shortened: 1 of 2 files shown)


s3/Cargo.toml
```diff
@@ -35,15 +35,15 @@ log = "0.4"
...
+reqwest = { version = "0.11", default-features = false, features = ["json", "stream"], optional = true }
 serde = "1"
 serde_derive = "1"
 serde-xml-rs = "0.4"
 sha2 = "0.9"
 anyhow = "1.0"
 surf = { version = "2", optional = true, default-features = false, features = ["hyper-client"] }
-# Waiting on h2 compat
-tokio = { version = "0.2", features = ["stream"], optional = true, default-features = false }
+tokio = { version = "1.0.1", features = ["io-util"], optional = true, default-features = false }
```
Jasperav, line 45: Maybe it is better to depend on tokio = "1.0" to include bugfixes (and to be consistent with the other dependencies, none depend on a specific version)

nightmared: Do note that this is already provided because of the default cargo semantic (except if the application linking to rust-s3 requires **exactly** tokio 1.0.0): https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#caret-requirements. So we are getting bugfixes, but only in the 1.0.x range. However, we probably want to get upgrades to 1.x instead of just 1.0.x, so that's a fair remark and I will fix it.

durch:
@nightmared thanks for the PR, it should all be going out over the weekend

durch:
published in `0.27.0-beta9`

Issue #164, opened 2021-01-22
Error Handling in 0.27


milenkovicm:
**Is your feature request related to a problem? Please describe.**

Handling errors with `0.27` gets much harder as methods are returning anyhow trait. It is a step back from `0.26` which had dedicated error(s). 

As `rust-s3` is a library change to `anyhow`  will make harder for users to handle errors.

**Describe the solution you'd like**

Have you considering exposing `thiserror` instead of `anyhow` ? 
Even `anyhow` documentation says 

> Comparison to thiserror
>
> Use Anyhow if you don't care what error type your functions return, you just want it to be easy. This is common in application code. Use thiserror if you are a library that wants to design your own dedicated error type(s) so that on failures the caller gets exactly the information that you choose

**Describe alternatives you've considered**

Other alternative would be to handle errors yourself but this would make no sense.

**Additional context**

Have no motivation to get into flame wars about library, this is just a suggestion for your consideration.

nightmared:
Hey ! I've being giving this a look today, would you mind tell me what you think of this?
https://github.com/durch/rust-s3/compare/dev...nightmared:thiserror

Also, before opening a PR, I think it would be useful to know if this is a wanted feature (Maybe @durch would prefer keeping `anyhow`?)

As an important note, this is not API compatible with the current 0.27 branch but I guess this is acceptable in a 0.x release.
Also, I fixed the two clippy 1.49 issues (one of them yields code that I actually find less legible that before, but oh well let's not upset clippy!).
finally, I added the sync tests to the Makefile so that it is executed when calling `make ci`.

milenkovicm:
Hi @nightmared, 
it looks good to me but it's up to @durch to decide.

Also, I'd recommend you to split the pr to two commits, one addressing thiserror other addressing clippy

durch:
@milenkovicm, @nightmared thank you both for the discussion, `anyhow` was mostly a spur of the moment decision. There has been a lot of development around error handling, and I'm sort of still stuck in `error_chain` days. I'm very open to a PR or additional discussion as to how we'd want to handle errors in general.

milenkovicm:
@nightmared did a decent job with on this issue in his branch, maybe that could be a good starting point to address this issue.

slckl:
Right now `0.27` feels like a downgrade of sorts with loss of custom error types.
`thiserror` would be strongly preferable to `anyhow` in error handling.

is there maybe something that can be helped with to get #189 over the line?

[...]

PR #165, opened 2021-01-28
[ci] pass checks introduced in clippy 1.49

nightmared:
Following https://github.com/durch/rust-s3/issues/164#issuecomment-768913909, this should fix errors when using clippy 1.49.


nightmared:
Fixed in https://github.com/durch/rust-s3/commit/3662f0c2ac2c7d9fe88287023e6e40ddd49a9550.

PR #166, opened 2021-01-30
Use thiserror to manage errors in rust-s3

nightmared:
Following discussion at #164, this PR replaces `anyhow` with `thiserror` to make errors more explicit, comparable, and so on. The goal is for consumers of this library to be able to extract information about errors without having to downcast anyhow errors, and thus to have more granularity to filter the errors and/or understand the issue at hand.

As an important note, this is not API compatible with the current 0.27 branch but I guess this is acceptable in a beta 0.x release.
I also added the sync tests to the Makefile so that it is executed when calling `make ci`.

### DIFF
--- a/aws-creds/Cargo.toml
+++ b/aws-creds/Cargo.toml
@@ -15,7 +15,7 @@ name = "awscreds"
 path = "src/lib.rs"
 
 [dependencies]
-anyhow = "1.0"
+thiserror = "1.0"
 dirs = "3"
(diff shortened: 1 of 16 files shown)


nightmared:
Rebased to solve conflits.

nightmared:
Considering this was closed, does this mean you expect me to reopen this PR against master, or that you don't want this feature?

Issue #178, opened 2021-03-25
S3 throws a duplicate error when fetching tags for an s3 object with multiple tags


feikesteenbergen:
**Describe the bug**
S3 throws a duplicate error when fetching tags for an s3 object with multiple tags

**To Reproduce**
- create s3 object

```rust
use s3::{creds::Credentials, Bucket, Region};

fn main() {
    let domain = "www.example.com";
    let credentials = Credentials::default().unwrap();
    let bucket = Bucket::new(domain, Region::UsEast1, credentials).unwrap();

    println!("{:#?}", bucket.get_object_tagging_blocking("ab.tgz"));
}
```

## a single tag
- Add the tag `tag1=value1` to the s3 object
- This is fine, expected behavior
```console
[...]
Ok(
    (
        Some(
            Tagging {
                tag_set: [
                    Tag {
                        kvpair: KVPair {
                            key: "tag1",
                            value: "value1",
                        },
                    },
                ],
            },
        ),
        200,
    ),
)
```

## two tags
- Add the tag `tag1=value1` to the s3 object
- Add the tag `tag2=value2` to the s3 object
- This is not fine, I expect to be able to receive something I can iterate over
```console
[...]
Err(
    S3Error {
        description: Some(
            "custom: duplicate field `Tag`",
        ),
        data: None,
        source: Some(
            Xml(
                Custom {
                    field: "duplicate field `Tag`",
                },
            ),
        ),
    },
)
```

**Expected behavior**
I would expect to be able to iterate over the tags. Instead I get a duplicate error.

**Environment**
 - Rust version: `1.50`
 - lib version [e.g. `0.26.4`]

durch:
@feikesteenbergen thanks for raising this, and sorry for the wait. This indeed does not look right, let me see what I can do

durch:
So this comes down to xml and serde and Rust. All in all a pretty disappointing story. TLDR serde can't really handle XML, I've switched to DOM for xml parsing and so far so good, fix will be published as `0.27.0-rc3`

feikesteenbergen:
@durch thanks for fixing this and working on this crate.

PR #186, opened 2021-04-09
Add Tag impl with key and value methods

### LINKED ISSUE #185: Tag `key()` and `value()` methods missing from 0.27.0-rc3
**Describe the bug**
There is no way to access `key` or `value` since they are private fields in the `Tag` struct and the `key()` and `value()` methods seem to have been removed.

**To Reproduce**
```rust
        let (tags, _code) = bucket.get_object_tagging(&some_object).await?;
        for tag in tags {
            let (k, v) = (tag.key(), tag.value())
        }
```

**Expected behavior**
I should be able to access `key` and `value` in the Tag struct as either public fields or a public method

**Environment**
 - Rust version: Any
 - lib version `0.27.0-rc3`

krlohnes:
There was no way to access these from the object in the crate, so I
added them back.

Fixes #185

### DIFF
--- a/s3/src/bucket.rs
+++ b/s3/src/bucket.rs
@@ -55,13 +55,13 @@ pub struct Tag {
 }
 
 impl Tag {
-   pub fn key(&self) -> String {
+    pub fn key(&self) -> String {
         self.key.to_owned()
-   }
+    }
 
-   pub fn value(&self) -> String {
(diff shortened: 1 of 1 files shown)


durch:
@krlohnes thank you, an oversight on my part, I've had to completely refactor tag handling due to #178. Published as `0.27.0-rc4`

PR #190, opened 2021-04-12
add custom headers on put_object requests

nightmared:
This is a repost of #155 (rebased on top of master), because I don't know why you closed the previous PR, and thus if this feature is wanted or not ?

I'm quite lazy, so I didn't want to upload objects with a presigned put (which allows for passing custom metadata). This means I had to be able to pass custom headers (in theory I only needed to add "x-amz-meta-*", but while we are at it, wa can as well let the user add any possible header, right ?). This is an attempt at solving this problem.
However, I do think that this PR is less than ideal because

* a builder pattern would probably prevent adding too many similar functions, and be better than this hacky patch
* we forward custom_headers to the s3 endpoint both for submitting parts & for initiating the upload, without distinguishing both (in the event of a multi-parts upload of course, otherwise there is no such dichotomy because there is only a single request). However adding two arguments would be tedious and there is no way to distinguish the headers targeting the parts from those targeting the InitiateMultipartUpload query. So I decided to send the same headers for both.

Anyway, do you think something like that would be valuable ?

### DIFF
--- a/s3/src/bucket.rs
+++ b/s3/src/bucket.rs
@@ -700,7 +700,7 @@ impl Bucket {
     /// // Blocking variant, generated with `blocking` feature in combination
     /// // with `tokio` or `async-std` features.
     /// #[cfg(feature = "blocking")]
-    /// let status_code = bucket.put_object_stream_blocking(&mut path, "/path")?;
+    /// let status_code = bucket.put_object_stream_blocking(&mut file, "/path")?;
     /// #
     /// # Ok(())
     /// # }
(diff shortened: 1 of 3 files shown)


durch:
@nightmared once more thanks for being patient, I think I've tried merging this in but ran into some sync/async trouble, once the `0.27.0` is out, I'll give it one more go, I do think it would be useful/valuable to have it in

durch:
@nightmared closing for now, not sure how badly you need something like this, I'd be happy to take a stab at it eventually

nightmared:
No problem, I can work around it by making a presigned put.

PR #195, opened 2021-05-04
Implement presign_delete

beingflo:
This PR implements `presign_delete`, mirroring `presign_get` exactly. I'm not exactly certain that the S3 spec intends for a presigned DELETE but rusoto seems to implement it 🤷‍♂️ .
This is tested against Exoscale and B2. For Exoscale it works as expected, B2 interestingly returns a CORS error if a DELETE is attempted on the URL after the expiration.

Note, this is was a quick fix and I didn't have a deep look at this code, so proceed accordingly.

### DIFF
--- a/s3/src/bucket.rs
+++ b/s3/src/bucket.rs
@@ -168,6 +168,29 @@ impl Bucket {
         );
         request.presigned()
     }
+
+    /// Get a presigned url for deleting object on a given path
+    ///
+    /// # Example:
+    ///
+    /// ```no_run
+    /// use s3::bucket::Bucket;
+    /// use s3::creds::Credentials;
+    ///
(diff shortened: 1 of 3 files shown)


durch:
@beingflo thank you for the contribution, it went under my radar, will review and merge promptly

dhbradshaw:
This and another issue are evidence that this code gets run successfully against Backblaze B2.

When I searched issues and code / documentation for Backblaze there were no hits.  I would have rejected the package if I hadn't thought to search for B2 instead of Backblaze.  So this comment is for the next person who wants to know if this works with Backblaze but doesn't search for B2.  It does!

durch:
@dhbradshaw I'm glad it does, it will be easier for other so find now! Thank you :)

PR #200, opened 2021-06-14
fix ECS instance detection when loading credentials from instance metadata

Ten0:


durch:
@Ten0 thank you for the contribution, I'll make sure to merge and publish a new version promptly

Ten0:
Note: we're currently using the fork in production, so this works.

Ten0:
up

Ten0:
up

PR #210, opened 2021-08-01
Exposing the multipart-upload api to users

johanot:
`list_multiparts_uploads` and `abort_upload` are already public functions.

But for client that support client-side-chunking, I currently have to assemble all chunks in a my application, only have them later re-chunked by rust-s3.

It should be possible for clients to make use of multipart upload directly, passing the relevant multipart state for each chunk uploaded.

Still draft: because docs (and possibly tests?) are missing.


durch:
@johanot Could you take a look at the conflict, I'd still like to get this merged, it slipped my mind, apologies for that

johanot:
@durch Done!

durch:
@johanot Amazing, thank you :)

PR #211, opened 2021-08-06
Implement CopyObject bucket operation

johanot:
Copy across different buckets is not exposed, purposefully, in order to not make the API too complex on the first impl.


durch:
@johanot Thanks for the contribution!

PR #214, opened 2021-09-16
Add a request timeout to credentials HTTP GETs

pkhuong:
Attohttpc sets connection and read timeouts by default.  However, by
default, there is no timeout on writing the request, nor any
end-to-end timeout.  Let applications impose one if they want to.


durch:
@pkhuong ditto for this one, great stuff!

PR #215, opened 2021-09-16
bucket: add optional HTTP timeout for all bucket operations

pkhuong:
And apply the new field in the attohttpc (blocking) and reqwest
backends.  Surf does not expose such an option, but it's also
not as useful in async code: the caller can await with timeout.


durch:
@pkhuong thank you!

Issue #162, opened 2021-01-13
Does not work with Minio


fredyflores96:
**Describe the bug**
I am not sure if this crate supports connecting to minio but I assume it may have been intended to since minio is on the the keywords listed when I look up the library. I am having lots of trouble making any requests to minio to create buckets or access any files. 

**To Reproduce**
Try to connect to any local minio pod and you will end up with XMinioUnknownAPIRequest responses. 

**Expected behavior**
I would expect it to support minio since minio can use the aws s3 cli to make its resquests. 

**Environment**
 - 1.44.0
 - 0.26.4

**Additional context**
Please let me know if I am just incorrectly trying to make calls to minio or if I am even supposed to try to use this library to try to communicate to minion.

some sample code I tried to use to connect 

```
struct Storage {
    name: String,
    region: Region,
    credentials: Credentials,
    bucket: String,
    location_supported: bool,
}

#[tokio::main]
async fn main()-> Result<(), S3Error> {
    let minio = Storage {
        name: "minio".into(),
        region: Region::Custom {
            region: "us-east-1".into(),
            endpoint: "http://localhost:9000".into(),
        },
        credentials: Credentials::from_profile(Some("minio-local"))?,
        bucket: "rust-s3".to_string(),
        location_supported: false,
    };

    for backend in vec![minio] {
        println!("Running {}", backend.name);
        let config = BucketConfiguration::default();
        let create_bucket_response = Bucket::create(&backend.bucket, backend.region, backend.credentials, config).await?;

        println!("{:#?}",create_bucket_response.response_text);
    }
    Ok(())
}
```

---this is the response when I print the response text:

`<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n<Error><Code>XMinioUnknownAPIRequest</Code><Message>Unknown API request at /</Message><Resource>/</Resource><RequestId></RequestId><HostId>780ff3b5-f965-4153-ad4d-893237116cb2</HostId></Error>`

Thanks!

durch:
Hi @fredyflores96 it should work with `minio` the gotcha is to use `path_style` buckets - [Bucket::new_with_path_style](https://docs.rs/rust-s3/0.26.4/s3/bucket/struct.Bucket.html#method.new_with_path_style)

fredyflores96:
[...]

PR #217, opened 2021-10-04
Add some commented example for Minio

yageek:
With Minio, the bucket name is part of the path of the requests, so `Bucket` instances would need to be instantiated using `new_with_path_style`.

This add some smalls commented lines inside the referenced example (as suggested in #162) to emphase it.

This would help starters to save time getting everything running up.


durch:
@yageek thank you!

PR #219, opened 2021-10-11
Fix crates.io badge, update links and various formatting

atouchet:


durch:
@atouchet thank for the contribution!

PR #220, opened 2021-10-11
aws-creds: default to a 30-second request timeout

pkhuong:
This is technically a breaking change, but no (infinite) timeout is a bad default, and incentivises libraries to override the global timeout at all sorts of different values. We have been running with a 30-second timeout in production for more than a month, without any problem: in a healthy state, credentials requests return in milliseconds. That seems like a reasonable, if conservative, default.


durch:
Anything is better then Inf :)

PR #221, opened 2021-10-11
s3/bucket: default to 30-second request timeout

pkhuong:
No (infinite) timeout is a bad default. It's easy for libraries to override their bucket objects' timeout, but we should probably default to *some* timeout, rather than letting requests remain stuck forever (up to TCP timeout).

I'm not 100% convinced this change is a good idea: large uploads or downloads can last a more than 30 seconds... On the other hand, it's easier to observe a timeout and relax the maximum duration, compared to realising that a program is stuck because there's no timeout on S3 calls. What do you think?


durch:
Thank you! Good one

PR #223, opened 2021-10-11
aws-creds: simplify and deduplicate `is_ec2` check

pkhuong:
1. Call `is_ec2` in `Credentials::from_instance_metadata` instead of
   duplicating a slightly different logic.

2. Use `str::starts_with` when that's what we mean.

The only observable difference is that `::from_instance_metadata` will
now accept a machine as being on ec2 if either `/sys/hypervisor/uuid`
or `/sys/class/dmi/id/board_vendor` match, while the previous version
only did when *both* files matched.  That seems reasonable enough.

### DIFF
--- a/aws-creds/src/credentials.rs
+++ b/aws-creds/src/credentials.rs
@@ -278,13 +278,7 @@ impl Credentials {
                     .json()?
             }
             Err(_) => {
-                if !(std::fs::read_to_string("/sys/hypervisor/uuid")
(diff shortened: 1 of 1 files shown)


pkhuong:
Assuages clippy.

durch (approved):

durch:
Nice!

durch:
Ported out of PR as, I did a similar tweak already

PR #226, opened 2021-10-12
Make minidom dependency optional

jplatte:
This one function pulls in both minidom and quick-xml, and we have never used it. Would be nice to be able to get rid of those extra dependencies.


durch (approved):

durch:
Thank you @jplatte !

PR #227, opened 2021-10-12
Import futures-io crate directly instead of through futures

jplatte:
This can heavily reduce dependency count for applications that don't otherwise depend on some of the `futures-*` crates.

### DIFF
--- a/s3/Cargo.toml
+++ b/s3/Cargo.toml
@@ -27,7 +27,7 @@ aws-region = "0.23"
 base64 = "0.13.0"
 cfg-if = "1"
 chrono = "0.4"
-futures = { version = "0.3", optional = true }
+futures-io = { version = "0.3", optional = true }
 hex = "0.4"
 hmac = "0.11"
 http = "0.2"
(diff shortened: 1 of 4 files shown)


s3/Cargo.toml
```diff
@@ -50,8 +50,8 @@ minidom = "0.13"
 block_on_proc = { version = "0.2", optional = true }
 
 [features]
-with-tokio = ["reqwest", "tokio", "futures", "tokio/fs"]
-with-async-std = ["async-std", "surf", "futures"]
+with-tokio = ["reqwest", "tokio", "futures-io", "tokio/fs"]
```
asonix, line 53: tokio doesn't need the futures-io feature

durch (approved):

durch:
@jplatte looks like this blows up `presign` tests, as the slashes in paths do not get encoded, holding off on merging for now

durch:
@jplatte unfortunatley this breaks async-std as we also need the `AsyncReadExt` trait which is not in `futures_io`

jplatte:
Maybe that's a difference between `master` and `develop`. Should I create a followup PR?

jplatte:
Oh, I see what's wrong. Will open another PR.

PR #230, opened 2021-10-31
Use async write for tokio,async-std

asonix:
Rather than passing std::io::Write to async methods, define separate tokio/async-std/std versions of the `response_data_to_writer` method on `Request` to accept the proper sync/async Write trait for the environment


durch (approved):

durch:
Another nice one! Thank you

PR #231, opened 2021-10-31
Make attohttpc optional for credentials

asonix:
If an application is simply constructing a Credentials struct from existing values for `rust-s3`, it doesn't need an http client bundled into the library. This change makes the http client in aws-creds optional behind a feature flag, but still enabled by default.


durch (approved):

durch:
Good stuff, thank you!

PR #233, opened 2021-11-03
Make all fail-on-err variations run the same check

asonix:
All 3 provided client implementations contain an `is_success` method for status codes, and the behavior of all 3 of those methods is the same, so this standardizes the behavior between the clients.


durch:
@asonix nice one, thank you!

Issue #188, opened 2021-04-10
Reuse creation of Client


Jasperav:
Currently a new Client is created for each time a request is made, e.g.: https://github.com/durch/rust-s3/blob/69e3efe3a7219bc9d2aab8535b74265642e22d45/s3/src/request.rs#L77

The documentation of Client tells us that the Client should be reused: https://github.com/seanmonstar/reqwest/blob/7d8c32784e543b1a2b9536a70db466c0250876c2/src/async_impl/client.rs#L54

It is better for performance to cache. What do you think about this?

ahrens:
I think this relates to (perhaps duplicates?) https://github.com/durch/rust-s3/issues/152.

durch:
Due to the way the request is currently architected, I don't see it happening any time soon

Issue #152, opened 2020-12-21
Re-use the same HTTP client instance 


glebpom:
**Is your feature request related to a problem? Please describe.**
Re-creating reqwest/surf instance is not effective, since the connection pool cannot be re-used.

**Describe the solution you'd like**
Request implementation may be refactored, so that Client instance is stored in the particular implementation struct, and move `command` and `path to trait functions arguments

durch:
@glebpom thanks for the call out, a few versions ago I switched this back to spawning a dedicated client due to issues with tests. I'm in general pro this approach, its just a hassle to rework the tests, I'll have some more free time in the next few weeks so I might revisit this. I'm always open to PRs :)

Ten0:
It looks like this could even use a single `reqwest::Client` in a `lazy_static`, given that all requests are made to the same endpoints. This may allow e.g. reusing SSL sessions across several objects, so there's a good chance this would be the best approach.
Also, this may solve your issue with tests.

durch:
Same as #234, while we could add it to `lazy_static` its a bit too magical for my taste, it would be ideal to allow ppl to bring their own client, something I'm thinking through

Issue #212, opened 2021-08-13
Async causes OOM when transferring from faster sources than destinations


sempervictus:
**Describe the bug**
In a use case where messages are being streamed off a queue and dispatched via this library, `async` calls result in unbounded memory growth, file descriptor exhaustion, and absurd object commit measurements. So a Tokio main calling the following:
```rust
async fn send_s3(msg: &Message<'_>, bucket: Bucket) -> Result<String,Box<dyn Error> > {
    let s3_timer = METRIC_INGESTOR_S3_UPLOAD
        .with_label_values(&["all"])
        .start_timer();
    let path = s3_msg_path(msg.clone());
    // println!("S3 Path: {}", &path);
    let (_, code) = bucket.put_object(&path, msg.source.as_bytes()).await.unwrap();
    let upload = "S3 upload result: ".to_string() + &code.to_string();
    s3_timer.observe_duration();
    Ok(upload)
}
```
results in the observed duration being ~9s, memory growing non-stop until OOM, and an 8c system uploads to a single minio @ ~25msg/s with a lot of jitter on the graphs.
Replacing `async` with a `rayon::ThreadPool` and the `send_s3` rewritten as
```rust
fn send_s3(msg: &Message<'_>, bucket: Bucket) -> Result<String,Box<dyn Error> > {
    let s3_timer = METRIC_INGESTOR_S3_UPLOAD
        .with_label_values(&["all"])
        .start_timer();
    let path = s3_msg_path(msg.clone());
    // println!("S3 Path: {}", &path);
    let (_, code) = bucket.put_object_blocking(&path, msg.source.as_bytes()).unwrap();
    let upload = "S3 upload result: ".to_string() + &code.to_string();
    s3_timer.observe_duration();
    Ok(upload)
}
```
called synchronously from the threads, i get no memory growth, and message rates are a flat ~27/s with an observed duration of ~0.9s (still awful).

**Expected behavior**
Async do not await a return for 9s, and dont keep consuming off the queue when they're still not done with all of their dispatch. I think the ideal here would be to fire-and-forget the S3 message in that once the PUT/POST is sent, the source data buffer should be freed unless referenced elsewhere. Otherwise as the main thread eats off the MQ and dispatches async S3 puts, the memory used by those messages doesn't free and any reasonably long-running binary crashes.

**Environment**
 - Rust version: [e.g. `1.54`]
 - lib version [e.g. `0.26`]

**Additional context**
[...]

PR #234, opened 2021-11-03
Make request methods generic over a new Client type

asonix:
This has a few benefits:
1. Consumers of this crate are now capable of bringing their own HTTP Clients (so long as they produce futures that are Send)
2. Consumers can enable runtime support (tokio v async-std v sync) without being forced into a specific Client choice
3. Clients are no longer constructed for each request and can be re-used for multiple requests, meaning we can make use of a client's built in connection pooling instead of building new clients per request

Drawbacks:
1. Every method on Bucket now expects a &Client argument. This could be worked around by storing the client inside a bucket, but I personally prefer to keep them separate

### DIFF
--- a/s3/Cargo.toml
+++ b/s3/Cargo.toml
@@ -45,7 +45,7 @@ serde-xml-rs = "0.5"
 sha2 = "0.10"
 thiserror = "1"
 surf = { version = "2", optional = true, default-features = false, features = ["h1-client-rustls"] }
-tokio = { version = "1", features = ["io-util"], optional = true, default-features = false }
+tokio = { version = "1", features = ["io-util" ], optional = true, default-features = false }
 tokio-stream = { version = "0.1", optional = true }
 url = "2"
 minidom = { version = "0.14", optional = true }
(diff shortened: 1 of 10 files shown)


asonix:
re-using the same client between requests potentially helps this problem: https://github.com/durch/rust-s3/issues/212

Parth:
Also 
+ fixes: #152
+ fixes: #188

Very happy to see this being worked on

durch:
@asonix thanks for the PR, my bandwidth is a bit limited, but I'll try and get this merged soon-ish

Parth:
Thanks @durch

asonix:
Was this intended to be closed or was that a result of deleting the target branch

durch:
Its the target branch, lets remedy that

Parth:
eagerly awaiting this being merged :crossed_fingers:

durch:
Closing for now, I'm not, branches have diverged too far, with all the features and different async backends this would be a tall order merge :/

PR #236, opened 2021-11-11
Lumeohq presign fix

durch:
Continuation of #225

### DIFF
--- a/s3/src/bucket.rs
+++ b/s3/src/bucket.rs
@@ -2264,7 +2264,6 @@ mod test {
     }
 
     #[test]
-    #[ignore]
     fn test_presign_put() {
         let s3_path = "/test/test.file";
         let bucket = test_aws_bucket();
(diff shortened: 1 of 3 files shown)


durch:
@jplatte looks like this blows up `presign` tests, as the slashes in paths do not get encoded, holding off on merging for now

jplatte:
You have some string comparisons in those tests where you explicitly test whether `%2F` or `/` is encoded. It's obvious that they would change, but have you tested whether this works with actual S3 server implementations?

durch:
> You have some string comparisons in those tests where you explicitly test whether `%2F` or `/` is encoded. It's obvious that they would no longer pass, but have you tested whether this works with actual S3 server implementations?

No, good point, lets see whats what in reality :)

jplatte:
@durch Any update on this?

durch:
@jplatte just looking into this again, thanks for the patience

jplatte:
Looks like this wasn't included in v0.28.0 or v0.28.1, was that intentional?

jplatte:
@durch Was this removed again intentionally? Looks like the commit you merged into develop never made it to master, and now you removed it from develop too 🙁

PR #237, opened 2021-11-11
Import futures-* crates directly instead of through futures

jplatte:
Was partially broken due to #230 introducing new imports from `futures`, and partially due to me missing that it should have been `futures_util` in one place instead in #227.


jplatte:
I see you reverted #227. I've updated this PR to re-apply it such that it doesn't break async-std.

PR #242, opened 2021-11-29
fix #241

### LINKED ISSUE #241: panic for https://github.com/durch/rust-s3/blob/master/s3/src/surf_request.rs(line 63)
https://github.com/durch/rust-s3/blob/master/s3/src/surf_request.rs     (line 63)
```rust
 let response = request.send().await.unwrap();
```

zhuxiujia:
fix #241


durch:
@zhuxiujia thanks for the fix!

Issue #246, opened 2022-01-24
Build fails after rust version 1.58.1 form 01-20-2022


Meph1587:
**Describe the bug**
After updating rust to `1.58.1` with `rustup update` build of this crate fails

**To Reproduce**
make project with `rust-s3: 0.28` as dependency and compile with new version of rust

**Expected behavior**
Should compile 

**Environment**
 - Rust version: [e.g. `1.58.1`]
 - lib version [e.g. `0.28`]

**Additional context**
``` bash
 Compiling rust-s3 v0.28.0
error: future cannot be sent between threads safely
   --> /Users/u/.cargo/registry/src/github.com-1ecc6299db9ec823/rust-s3-0.28.0/src/request.rs:47:50
    |
47  |       async fn response(&self) -> Result<Response> {
    |  __________________________________________________^
48  | |         // Build headers
49  | |         let headers = match self.headers() {
50  | |             Ok(headers) => headers,
...   |
102 | |         Ok(response)
103 | |     }
    | |_____^ future created by async block is not `Send`
    |
    = help: the trait `Sync` is not implemented for `core::fmt::Opaque`
note: future is not `Send` as this value is used across an await
   --> /Users/u/.cargo/registry/src/github.com-1ecc6299db9ec823/rust-s3-0.28.0/src/request.rs:98:32
    |
95  |               return Err(anyhow!(
    |  ________________________-
96  | |                 "Request failed with code {}\n{}",
97  | |                 response.status().as_u16(),
98  | |                 response.text().await?
    | |                                ^^^^^^ await occurs here, with `$crate::__export::format_args!($($arg)*)` maybe used later
99  | |             ));
    | |             -
    | |             |
    | |_____________has type `ArgumentV1<'_>` which is not `Send`
    |               `$crate::__export::format_args!($($arg)*)` is later dropped here
    = note: required for the cast to the object type `dyn Future<Output = Result<reqwest::Response, anyhow::Error>> + Send`

error: could not compile `rust-s3` due to previous error
warning: build failed, waiting for other jobs to finish...
error: build failed
```

jplatte:
This seems like a regression in rustc, has somebody opened an issue on [rust-lang/rust](https://github.com/rust-lang/rust)?

jplatte:
I think this was fixed a few nightly releases back.

durch:
@jplatte thanks for being on top of this!

PR #247, opened 2022-01-24
Fix build error - #246

Meph1587:
Fixes the build error from #246 by moving request parsing outside of `anyhow!()` and formatting


durch:
@Meph1587 could you pls handle these few failing clips checks? Thank you!

Meph1587:
Sure done now

Issue #163, opened 2021-01-13
aws-creds's `Credentials` initilaized with `from_instance_metadata` are broken after 6 hours.


Ten0:
**Describe the bug**
aws-creds's `Credentials` initilaized with `from_instance_metadata` are broken after 6 hours, because they expire, as explained here:
https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-iam-roles.html

**To Reproduce**
Initialize `Credentials` with `from_instance_metadata`. Wait 6 hours. Attempt an action that requires these credentials to be valid.

**Expected behavior**
*still works*

**Environment**
 - Rust version: latest stable (`1.49`)
 - lib version: aws-creds `0.26.0`

**Solving**
`Credentials` should probably be an structure containing either:
 - constant credentials
 - that they should be fetched from the instance, that the current are *these*, and that they expire at `Expiration`. (`Expiration` is one of the keys returned by the call to `169.254.170.2$AWS_CONTAINER_CREDENTIALS_RELATIVE_URI`).

In the second case, they should be automatically re-fetched whenever they expire.

durch:
@Ten0 another great catch, I'll look into automatically refreshing these, that is the expected behavior for sure.

durch:
@Ten0 sorry for going AWOL on this, work/life asserted themselves, will try and pick it up next week

Ten0:
No worries, thanks for your dedication :)

kennetpostigo:
Question, does this mean accessing buckets after 6 hours will not work and you will need to re-create and populate the bucket every 6 hours?

Ten0:
Not "bucket" in the sense of "AWS bucket", but `Bucket` in the sense [`aws_s3::Bucket`](https://docs.rs/rust-s3/0.26.4/s3/bucket/struct.Bucket.html). The struct has to be rebuilt, because the credentials stored inside become invalid after 6 hours so it just stops working.

Ten0:
What is this fixed by?

durch:
It wasn't, I have no idea how this got closed, I probably fat fingered it :/

PR #249, opened 2022-01-31
Allow access to credentials expiration

Ten0:
Following up on https://github.com/durch/rust-s3/issues/163#issuecomment-907420563.

This is a first step towards fixing #163, and it at least allows to work around it correctly.

Note: this is a breaking change for explicit constructions/destructurations of `Credentials`, so it will require a new major (`0.n+1`) version of aws-creds.


Ten0:
done

durch:
Cherry picked locally, will be published in the next release, thank you!

PR #250, opened 2022-02-01
Migrate from `chrono` to `time` to solve RUSTSEC-2020-0159

MarijnS95:
`chrono` still hasn't found a solution to [RUSTSEC-2020-0159] whereas `time` already solved its vulnerability to [RUSTSEC-2020-0071] by hiding the affected functionality behind a cfg flag (not to be confused with a `feature`, such `cfg`s can only be enabled through `RUSTFLAGS`).  At the same time `chrono` is a superset of `time` even though this crate hardly uses any functionality of it: only UTC time is needed which does not suffer from aforementioned local time vulnerabilities.

[RUSTSEC-2020-0071]: https://rustsec.org/advisories/RUSTSEC-2020-0071
[RUSTSEC-2020-0159]: https://rustsec.org/advisories/RUSTSEC-2020-0159

---

Note that this changes some public API which may or may not be desired?  `chrono` hasn't updated to the latest `time` either so even if there are helpers to go from `chrono` to `time` types seamlessly that won't work.  Perhaps going to and from Rust's `SystemTime` works but I haven't investigated.


durch:
@MarijnS95 thank you!

PR #254, opened 2022-02-14
Remove never-encode-slash feature

jplatte:
and make the right behavior the default.

This was merged in #236, but never made it to a release and now it's gone again on `develop` too. I'm hoping that was just an accident.


jplatte:
I can update the tests if you decide you want this in again, but the whole situation is a bit unclear to me for now so I'll wait.

durch:
@jplatte thanks for the catch, an omission on my part

PR #257, opened 2022-02-25
Providing more detailed error when the put API fails

castaneai:
Hi. I got the serde_xml error `custom: missing field Key` when using the upload API. However, the real cause was that the put API was failing.

Providing more detailed information when the put API fails will make it easier to understand the cause.


durch (approved):
Thanks @castaneai!

PR #262, opened 2022-03-16
Add custom_queries to command PresignGet

qwfy:
Added a field `custom_queries` to command `PresignGet`.

The motivation is to be able to set the `response-content-disposition` query parameter (causing the server to respond with the `Content-Disposition` header set to the specified value) to enable the browser to save the presigned url as a file using the specified file name.


durch:
@qwfy thanks for the PR, will be released in 0.31

PR #265, opened 2022-03-28
s3: Bump hmac to 0.12, sha2 to 0.10 and minidom to 0.14

Quartenia:


durch:
@Quartenia thanks for the PR, will be released in 0.31.

PR #267, opened 2022-04-09
make some fields in struct ListBucketResult optional

mateuszkj:
**Describe the bug**
Some S3 self-hosted servers do not returns all fields during listing files. This merge request makes that fields optional.

I run with this problem with S3 servers:
 - https://github.com/datenlord/s3-server (also written in Rust)
 - https://github.com/chrislusf/seaweedfs
 
 I tested only `ListObjectsV2` and `GetObjectRange` commands with above servers.

**Additional context**

I checked whats fields are required in `aws-sdk-rus` and there almost all are optional, even `key`. 
  - `ListObjectsV2Output` - https://github.com/awslabs/aws-sdk-rust/blob/main/sdk/s3/src/output.rs#L2366=
  - `Object` - https://github.com/awslabs/aws-sdk-rust/blob/main/sdk/s3/src/model.rs#L13517=

### DIFF
--- a/s3/src/serde_types.rs
+++ b/s3/src/serde_types.rs
@@ -13,7 +13,7 @@ pub struct InitiateMultipartUploadResponse {
 pub struct Owner {
     #[serde(rename = "DisplayName")]
     /// Object owner's name.
-    pub display_name: String,
+    pub display_name: Option<String>,
     #[serde(rename = "ID")]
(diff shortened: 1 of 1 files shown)


durch:
@mateuszkj thanks for the PR, will be released in 0.31.

PR #276, opened 2022-06-11
Set the patch version of `time`

hotate29:
The `time::format_description::well_known::Rfc2822` used in this crate
has been added since `time` version `0.3.6`. Therefore, Cargo.toml must set `0.3.6`.


durch:
Thanks for the contribution!

PR #277, opened 2022-06-23
Derive Serialize and Deserialize for Region and Credentials

fmckeogh:


durch:
@fmckeogh thank you for the contribution!

PR #281, opened 2022-07-26
Allow turning ResponseData into Vec<u8> without reallocating

### LINKED ISSUE #280: Regression: Can't obtain file body as Vec<u8> without reallocating since "Proper response type"
Since https://github.com/durch/rust-s3/commit/0484173489335ad2563dcbd43c269c4fd8275ce1 we can only access the file body as a slice in `ResponseData`.
This requires reallocating the entire file contents if needing to return it as a `Vec<u8>` (for backend-agnostic interface).

```rust
impl ResponseData {
    fn into_bytes(self) -> Vec<u8> {
        self.bytes
    }
}
```

**Additional context**
This is an regression encountered when updating from 0.31 to 0.32.

Ten0:
Fix #280

### DIFF
--- a/s3/src/request_trait.rs
+++ b/s3/src/request_trait.rs
@@ -32,6 +32,10 @@ impl ResponseData {
     pub fn status_code(&self) -> u16 {
         self.status_code
     }
+
+    pub fn into_bytes(self) -> Vec<u8> {
+        self.bytes
+    }
 }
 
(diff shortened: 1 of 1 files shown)


durch:
@Ten0 I've totally missed the PR, v0.32.3 provides the `Into<Vec<u8>>` for `ResponseData`

PR #284, opened 2022-07-29
Fix presigining

### LINKED ISSUE #283: Presigned links broken as of 0.32
**Describe the bug**
Presigned links are deemed invalid by AWS as of version 0.32.
They worked fine in version 0.31.

**To Reproduce**
Generate a presigned link in 0.32.
See the following error message back from AWS:
```xml
<Error>
<Code>InvalidToken</Code>
<Message>
The provided token is malformed or otherwise invalid.
</Message>
```

**Expected behavior**
*works*

**Environment**
 - Rust version: latest stable
 - lib version: `0.32.1`

**Additional context**

Looks like at least the `X-Amz-Security-Token` variable is not properly urlencoded in 0.32, while it does get properly urlencoded in previous versions.

Ten0:
FIX #283

Also remove a bunch of unnecessary intermediate string allocations

### DIFF
--- a/s3/src/signing.rs
+++ b/s3/src/signing.rs
@@ -84,7 +84,13 @@ pub fn canonical_query_string(uri: &Url) -> String {
     keyvalues.sort();
     let keyvalues: Vec<String> = keyvalues
         .iter()
-        .map(|(k, v)| format!("{}={}", uri_encode(k, true), uri_encode(v, true)))
+        .map(|(k, v)| {
+            format!(
+                "{}={}",
+                utf8_percent_encode(k, FRAGMENT_SLASH),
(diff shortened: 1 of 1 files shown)


s3/src/signing.rs
```diff
@@ -219,11 +224,16 @@ pub fn authorization_query_params_no_sig(
...
-        signed_headers = signed_headers_string
+        signed_headers = signed_headers,
     );
 
     if let Some(token) = token {
-        write!(query_params, "&X-Amz-Security-Token={}", token).expect("Could not write token");
+        write!(
+            query_params,
+            "&X-Amz-Security-Token={}",
+            utf8_percent_encode(token, FRAGMENT_SLASH)
```
Ten0, line 234: this is the actual fix

Issue #275, opened 2022-06-03
Get stream API


mihaigalos:
**Is your feature request related to a problem? Please describe.**
No.

**Describe the solution you'd like**
Using the `async` flavor, would like to have an API call to get a stream object (implementing trait `futures_core::stream::Stream`) for `GET`/`PUT` a file to a bucket.
This will then be used in a loop to write/read to/from a file and update a progress bar.

**Describe alternatives you've considered**
Alternatively, provide an API call to a observable callback with the current progress (percentage or transferred bytes), which clients can dependency-inject during `Bucket::new()`.
The callback would be called during io operations with the bucket.
Other suggestions are welcome.

**Additional context**
Consider the following code. This is how I would consume the stream for a `GET`. Similar for a `PUT`.

<details>
  <summary>Click to expand!</summary>

```rust
async fn get(input: &str, output: &str, bar: &mut WrappedBar) -> Result<(), ValidateError> {
    let parsed_address = ParsedAddress::parse_address(input, bar.silent);
    let (mut out, mut downloaded) = io::get_output(output, bar.silent);

    let res = reqwest::Client::new()
        .get(input)
        .header("Range", "bytes=".to_owned() + &downloaded.to_string() + "-")
        .header(
            reqwest::header::USER_AGENT,
            reqwest::header::HeaderValue::from_static(CLIENT_ID),
        )
        .basic_auth(parsed_address.username, Some(parsed_address.password))
        .send()
        .await
        .or(Err(format!("Failed to GET from {} to {}", &input, &output)))
        .unwrap();
    let total_size = downloaded + res.content_length().or(Some(0)).unwrap();

    bar.set_length(total_size);

    let mut stream = res.bytes_stream();
    while let Some(item) = stream.next().await {
        let chunk = item.or(Err(format!("Error while downloading."))).unwrap();
        out.write_all(&chunk)
            .or(Err(format!("Error while writing to output.")))
            .unwrap();
        let new = min(downloaded + (chunk.len() as u64), total_size);
        downloaded = new;
        bar.set_position(new);
    }

    bar.finish_download(&input, &output);
    Ok(())
}
```
</details>

durch:
As of `0.33.0-beta1` `get_object_stream` returns a `ResponseDataStream`, which has a `bytes()` method that is `Stream<Item = Bytes>`, I'd appreciate if you could test it out. I expect tokio backend works much better then async-std for tnis

mihaigalos:
[...]

PR #285, opened 2022-08-01
Request streaming

Janrupf:
This is a partial implementation of #275, allowing the streaming of download requests using async `Stream<Item = Result<Bytes, S3Error>>` streams.

There is no equivalent version for synchronous rust.


durch (approved):

durch:
@Janrupf nice, I like this one a lot

PR #287, opened 2022-08-03
Fix date and time parsing

### LINKED ISSUE #286: STS credentials are broken
**Describe the bug**
Trying to obtain STS credentials yields a serde error.

**To Reproduce**
Any use of `Credentials::from_sts()`.

**Expected behavior**
Credentials parse successfully

**Environment**
 - Rust version: 1.62.0
 - lib version: 0.32.2

**Additional context**
Serde error: `ParseIntError: invalid digit found in string`

Janrupf:
Fixes #286

**WARNING**: This is a possibly breaking change due to the use of a newtype wrapper in a public interface.

The newtype wrapper is completely transparent and derefs to its inner type, however, this could still break and require adding an `.into()`.

### DIFF
--- a/aws-creds/src/credentials.rs
+++ b/aws-creds/src/credentials.rs
@@ -1,10 +1,12 @@
 #![allow(dead_code)]
+
 use crate::error::CredentialsError;
 use ini::Ini;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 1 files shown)


durch (approved):

durch:
@Janrupf big thank you for the fix!

PR #290, opened 2022-08-16
fix: `get_object_tagging` for minio

not-matthias:
MinIO doesn't set a namespace in the response. They also don't call it `Values`, but `Value`. 
```xml
<?xml version="1.0" encoding="UTF-8"?>
<Tagging>
    <TagSet>
        <Tag>
            <Key>foo</Key>
            <Value>bar</Value>
        </Tag>
    </TagSet>
</Tagging>
```


durch:
@not-matthias thank you :)

PR #292, opened 2022-08-25
Fix StsResponseCredentials deserialization

Ten0:
As documented here, this field is rfc3339 as well (and `time` by default has chosen a random non-standard format):
https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRoleWithWebIdentity.html


Ten0:
oh ok that was already fixed, nevermind

PR #294, opened 2022-09-09
fix anonymous credentials example display

LesnyRumcajs:
Fixed anonymous credentials example that was outside the code area.

image


durch:
@LesnyRumcajs thank you :)

PR #296, opened 2022-09-22
feature: return headers and status from get_object

### LINKED ISSUE #213: get_object should also return the requests headers (content-type, etc) so you don't need to make two requests
get_object doesn't return the request information.

get_object should also return the same as head_object in one request.

**Describe alternatives you've considered**
N/a

**Additional context**
N/a

ozgrakkurt:
closes #213


durch:
@ozgrakkurt thanks for the PR, will close it, I've implemented it more generally on the ResponseData object itself

Issue #278, opened 2022-07-25
Don't depend on reqwest


tomkarw:
**Is your feature request related to a problem? Please describe.**
I recently tried to remove reqwest as a dependency from a project (using the lower-level hyper::Client which was already required instead). After doing so, I noticed that `reqwest` is still included in the dependency tree due to its use in `rust-s3`.

Using `reqwest` (which relies on `hyper` either way) seems a suitable choice for application code, but `hyper` itself might mean fewer dependencies for a projects like mine.

**Describe the solution you'd like**
Remove `reqwest` dependency and replace its usages with underlying `hyper` and `rustls`.

**Describe alternatives you've considered**
Leaving it as it is and living with extra dependencies.

**Additional context**
I'd be happy to implement this myself, just let me know if you'd be interested. If yes, I'd try to code it straight away.

durch:
@tomkarw I like the idea, reqwest is pretty heavy, it might be a lot of work tough, I’m not sure how hyper works with tokio, which is the feature that is pulling reqwest in. I think you can get rid of reqwest by using the `sync` feature with default features off. That should pull attohttpc instead of reqwest. 

Regarding I’d be happy to take a PR removing it

tomkarw:
I have some experience with doing exacltly the `reqwest -> hyper` transition. I also proposed very similar change to another project that suffered from the same issue https://github.com/open-telemetry/opentelemetry-rust/issues/843.

You can assing this issue to me and I'll work on it.

tomkarw:
It would seem that merging https://github.com/durch/rust-s3/pull/234 first, might simplify, or even remove the need for this work.

I'll wait and see how that PR proceeds before working on this.

durch:
I've given up on reusing the client for now, I'm gonna look into using pure hyper, but can't commit to any timelines

tomkarw:
You can let me pick this up if you want. I'll be happy to do so.

durch:
For sure, go for it!

tomkarw:
@durch I did it, but now that I see it, I'm not sure if it's worth it. Your call.

drewkett:
[...]

PR #299, opened 2022-09-29
refactor(s3): don't depend on reqwest

tomkarw:
Closes https://github.com/durch/rust-s3/issues/278.


durch:
This is wonderful, thank you!

PR #301, opened 2022-10-02
Fix #300: Stream sync get object responses

### LINKED ISSUE #300: get_object_stream doesn't stream data
**Describe the bug**
In sync mode, `get_object_stream` reads whole file into memory.  I'm working on a memory constrainted system and a large file (>1gb) leads to an OOM kill.

**To Reproduce**

**Expected behavior**

A reasonable buffer should be used for streaming.

**Environment**

**Additional context**

This seems very similar to https://github.com/durch/rust-s3/commit/a4fdbdec508a90e97fcd60deefa037bbd5c1b822

andrewbaxter:


durch:
@andrewbaxter thank you for the fix, I've been vaguely aware of it

andrewbaxter:
Thanks for the review and merge! :bow:

PR #307, opened 2022-10-10
added native-tls-vendored feature

veeso:
These features should allow using the vendored libssl.


durch:
Nice, I'll publish new versions to get rid of paths

PR #309, opened 2022-10-26
fix: presigned custom header order

zifeo:
According to https://docs.aws.amazon.com/general/latest/gr/sigv4-create-canonical-request.html, the header should be ordered to obtain a correct signature.


zifeo:
@durch Can you please give a review?

durch:
@zifeo looks good, thank you!

PR #321, opened 2022-12-14
Change credential refresh log from info to debug

Eric-Arellano:
The new credentials refresh feature is extremely useful to our project - thank you! However, we're finding our logs have lots of noise with a refresh every 30 minutes. While we could change our RUST_LOG env var, I suspect it's more useful for the typical user for this log to be at the debug level.


durch (approved):

durch:
Good stuff, thank you!

PR #324, opened 2023-01-18
Fix etag generation for small files

theCapypara:
The etag algorithm was incorrect for files where the chunk size was < 2. The algorithm re-hashed the hashes, which isn't what is supposed to happen.
This PR fixes the algorithm by not rehashing the digest for chunk size 1.


durch:
@theCapypara great stuff, thank you

PR #326, opened 2023-01-24
Finish migrate from serde-xml-rs to quick-xml

joaommartins:
Bumps aws-creds version to 0.35 to force republishing.

Adds a  clippy `should_implement_trait` skip, as Credentials::default is implemented as a member function and not as an `impl` of `std::default::Default`.


jayvdb (approved):

durch:
@joaommartins thank, fixed as part of 0.33 release

PR #328, opened 2023-01-24
Bump s3 deps to new aws-creds

joaommartins:
This PR bumps the s3 crate dependencies to the new proposed aws-creds version. It depends on #326 and will not build until that PR is merged and the new version of aws-creds is available on the remote repository.

Also changes the base64 encode due to a deprecation in the base64 crate.


jayvdb (approved):

durch:
@joaommartins thanks for raising, bumped as part of general hygine

PR #330, opened 2023-01-28
Update base64 dependency to 0.21 plus minor refactoring

mina86:
Update base64 dependency to the latest 0.21 version and while doing
that refactor some of the code where base64 is used.  The refactoring
has a few benefits:
- base64 encoding is now done into an on-stack buffer avoiding String
  heap allocation,
- Inserting Content-MD5 is now in a helper function avoiding code
  duplication,
- match is used rather than if-else-if chain so that self.command() is
  now called only once and
- there are now fewer ‘to_string’ calls reducing allocations.

### DIFF
--- a/s3/src/request/request_trait.rs
+++ b/s3/src/request/request_trait.rs
@@ -1,5 +1,6 @@
 use hmac::Mac;
 use std::collections::HashMap;
+use std::convert::TryInto;
 #[cfg(any(feature = "with-tokio", feature = "with-async-std"))]
 use std::pin::Pin;
 use time::format_description::well_known::Rfc2822;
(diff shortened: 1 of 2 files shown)


s3/src/request/request_trait.rs
```diff
@@ -475,33 +477,41 @@ pub trait Request {
...
-            let mut range = format!("bytes={}-", start);
-
-            if let Some(end) = end {
-                range.push_str(&end.to_string());
-            }
+        fn insert_content_md5(
+            headers: &mut HeaderMap,
+            content: &[u8],
+        ) -> Result<(), http::header::InvalidHeaderValue> {
+            const CONTENT_MD5: HeaderName = HeaderName::from_static("content-md5");
```
durch, line 484: Not sure about this `const` over here

durch:
@mina86 thanks for raising, bumped the dependencies as part of general hygine

s3/src/request/request_trait.rs
mina86: It gets rid of run-time processing of the header name.  HeaderName::from_static first compares the header to known standard headers and, if no known header is found, verifies the value.  With const this is done at compile-time.

PR #331, opened 2023-01-28
Minor compilation fixes

mina86:
Firstly, observe that aws-creds doesn’t compile without default
features enabled.  This is because Credentials::refresh method which
calls Credentials::default which is only defined if http-credentials
feature is enabled.  Change the code (and related tests) so that the
former method is present only if that feature is enabled.

Note that even with that, rust-s3 doesn’t compile without default
features enabled.  This is because rust-s3 depends on aws-creds with
the above bug present.  This will be fixed once the dependency is
updated.

Secondly, fix minor compilation warnings caused by conditional
compilation leading to some dead code.

### DIFF
--- a/aws-creds/src/credentials.rs
+++ b/aws-creds/src/credentials.rs
@@ -188,6 +188,7 @@ fn http_get(url: &str) -> attohttpc::Result<attohttpc::Response> {
 }
 
 impl Credentials {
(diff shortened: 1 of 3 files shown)


durch:
@mina86, thanks for raising, fixed in `0.33`

PR #334, opened 2023-02-22
Add support for StreamReader

ishiev:
Add support for [tokio_util](https://docs.rs/tokio-util/0.7.7/tokio_util/index.html)::[io](https://docs.rs/tokio-util/0.7.7/tokio_util/io/index.html)::[StreamReader](https://docs.rs/tokio-util/0.7.7/tokio_util/io/struct.StreamReader.html#) in ResponseDataStream.bytes


durch:
@ishiev had to make a few changes to make it work with all the feature combinations, great stuff in principle, thank you

PR #343, opened 2023-05-07
draft: use async await & async sync rwlock for credentials in bucket

### LINKED ISSUE #337: Bucket - write and read lock for credentials does not work well with `tokio`
**Describe the bug**

When `rust-s3` is used with `tokio` and in a concurrent setup, refreshing the credentials will result in a read or write race condition for the corresponding lock.

The credentials just use the system `RwLock`.

```
#[derive(Clone, Debug)]
pub struct Bucket {
    pub name: String,
    pub region: Region,
    pub credentials: Arc<RwLock<Credentials>>,
    pub extra_headers: HeaderMap,
    pub extra_query: Query,
    pub request_timeout: Option<Duration>,
    path_style: bool,
    listobjects_v2: bool,
}

impl Bucket {
    pub fn credentials_refresh(&self) -> Result<(), S3Error> {
        Ok(self
            .credentials
            .try_write()
            .map_err(|_| S3Error::WLCredentials)?
            .refresh()?)
    }
}
```

**To Reproduce**
Just make some concurrent requests read requests:

```
// cargo run --example tokio

use awscreds::Credentials;
use s3::error::S3Error;
use s3::{Bucket, Region};

#[tokio::main]
async fn main() -> Result<(), S3Error> {
    let bucket = Bucket::new(
        "rust-s3-test",
        "eu-central-1".parse()?,
        // Credentials are collected from environment, config, profile or instance metadata
        Credentials::default()?,
    )?;

    let s3_path = "test.file";
    let test = b"I'm going to S3!";

    let response_data = bucket.put_object(s3_path, test).await?;
    assert_eq!(response_data.status_code(), 200);

    for _ in 0..200 {
        let b = bucket.clone();
        tokio::spawn(async move {
            let response_data = b.get_object(s3_path).await;
            if let Err(e) = response_data {
                eprintln!("{}", e);
            }
        });
    }

    Ok(())
}
```

About 1 % of the requests result in a write error
```
Could not get Write lock on Credentials
```

About 0,5 % of the requests result in a read error
```
Could not get Read lock on Credentials
```

**Expected behavior**
No error should occur.

**Environment**
 - Rust version: rustc 1.67.0 (fc594f156 2023-01-24)
 - lib version master (bea733a4b73cb8dda45b04ccdaa60771508bcb25) or 0.33

**Possible Solution**
The method should be `async` and bucket should use `tokio::sync::RwLock` when `tokio` is used.

a-nickol:
fixes #337

but i am not quite sure about the architecture of the library, since the fix uses a tokio and async std struct for synchronization, this PR is not usable for usage with feature `sync-native-tls`.

### DIFF
--- a/s3/src/bucket.rs
+++ b/s3/src/bucket.rs
@@ -67,6 +67,11 @@ impl Tag {
     }
 }
 
+#[cfg(all(feature = "with-tokio", not(feature = "with-async-std")))]
+type AsyncRwLock<T> = tokio::sync::RwLock<T>;
+#[cfg(all(feature = "with-async-std", not(feature = "with-tokio")))]
+type AsyncRwLock<T> = async_std::sync::RwLock<T>;
+
 /// Instantiate an existing Bucket
 ///
 /// # Example
@@ -85,7 +90,7 @@ impl Tag {
 pub struct Bucket {
     pub name: String,
     pub region: Region,
-    pub credentials: Arc<tokio::sync::RwLock<Credentials>>,
+    pub credentials: Arc<AsyncRwLock<Credentials>>,
     pub extra_headers: HeaderMap,
     pub extra_query: Query,
     pub request_timeout: Option<Duration>,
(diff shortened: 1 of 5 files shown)


FHTMitchell:
Hi we just ran into this problem, would like to see this merged...

durch:
This is a good idea, implemented in 729791a30b63ae72890d43d4e4ed4f6ca6a28e4b

Issue #251, opened 2022-02-03
Add support for ListBuckets API


silvestrpredko:
**Is your feature request related to a problem? Please describe.**
I faced a problem that I couldn't check if the bucket exists or not. Also, it would be nice to get a list of buckets for other purposes.

**Describe the solution you'd like**
Would be nice to have an API where I could pass credentials and region, then it returns a list of buckets.
Actually, this request is in [S3](https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListBuckets.html#API_ListBuckets_RequestParameters) API.

**Describe alternatives you've considered**
For now, I have a workaround. Just try to get a list of objects for the bucket. If the request fails bucket doesn't exist.
I don't like this.)

PR #348, opened 2023-06-04
Implement a new `ListBuckets` command and its consumer API.

aalekhpatel07:
# Summary

- **Feature**: Two new APIs to list existing buckets (i.e. `Bucket::list_buckets`), and to check if a given bucket instance exists (i.e. `Bucket::exists`).
- **Bug fix**: CI was failing because of an unused `tls_connector` variable, so its been renamed to `_tls_connector`. Not too sure if this is the ideal long-term fix but its a band-aid for now.

~~- **Feature**: The `serde_types` now store `chrono::DateTime<chrono::Utc>` instead of String, wherever there was any datetime-like data to be stored.~~ Turns out we get different formats across different datetime fields, so we can't specialize it into one type, unless we add a deserializer that guesses the format and tries to convert any datetime strings into a `chrono::DateTime<chrono::Utc>` by trying parsing different formats (and that probably merits another PR).

## `Bucket::list_buckets(region, credentials)`
**Given a region, and credentials, list all the buckets that are visible.**
### Usage
```rust
use s3::{Bucket, BucketConfiguration};
use s3::creds::Credentials;
use s3::region::Region;
use anyhow::Result;

#[tokio::main]
async fn main() -> Result<()> {
    let region = Region::Custom {
      region: "eu-central-1".to_owned(),
      endpoint: "http://localhost:9000".to_owned()
    };
    let credentials = Credentials::default()?;
    
    // Async variant with `tokio` or `async-std` features
    let response = Bucket::list_buckets(region, credentials).await?;
    
    // `sync` feature will produce an identical method
    #[cfg(feature = "sync")]
    let response = Bucket::list_buckets(region, credentials)?;

    // Blocking variant, generated with `blocking` feature in combination
    // with `tokio` or `async-std` features.
    #[cfg(feature = "blocking")]
    let response = Bucket::list_buckets_blocking(region, credentials)?;

    let found_buckets = response.bucket_names().collect::<Vec<String>>();
    println!("found buckets: {:#?}", found_buckets);
    Ok(())
}
```

## `Bucket::exists`
**Determine whether the instantitated bucket exists.**

### Usage
```rust
use s3::{Bucket, BucketConfiguration};
use s3::creds::Credentials;
use s3::region::Region;
use anyhow::Result;

#[tokio::main]
async fn main() -> Result<()> {
    let bucket_name = "some-bucket-that-is-known-to-exist";
    let region = Region::Custom {
      region: "eu-central-1".to_owned(),
      endpoint: "http://localhost:9000".to_owned()
    };
    let credentials = Credentials::default()?;
    
    let bucket = Bucket::new(bucket_name, region, credentials)?;
    
    // Async variant with `tokio` or `async-std` features
    let exists = bucket.exists().await?;
    
    // `sync` feature will produce an identical method
    #[cfg(feature = "sync")]
    let exists = bucket.exists()?;

    // Blocking variant, generated with `blocking` feature in combination
    // with `tokio` or `async-std` features.
    #[cfg(feature = "blocking")]
    let exists = bucket.exists_blocking()?;
    
    assert_eq!(exists, true);
    Ok(())
}
```


aalekhpatel07:
haha. I didn't know when I issued this PR but this also closes #251.

durch:
@aalekhpatel07 Nice work, thank you!

aalekhpatel07:
Thank you for merging!

PR #349, opened 2023-06-05
removed debug print

pfz4:
this library logged a "CompletedMultipartUpload: ..." message to console. 
As this was the only active print statement in the entire project I concluded that it shouldn't be there and removed it.


durch:
Fixed in a previous PR

PR #353, opened 2023-06-23
Only send metadata on Initiate of multipart upload

### LINKED ISSUE #352: AWS returns `InvalidArgument` when multipart object includes metadata
**Describe the bug**
User-defined metadata -- header names with the `x-amz-meta-` prefix -- should only be sent with the `initiate_multipart_upload`. Assuming `Bucket::with_extra_headers` is the correct way to set metadata on the object, and since the headers are passed to every request involved in a multipart upload, AWS errors when it sees the metadata on the first chunk sent after the `initiate`.

**To Reproduce**
We're uploading very large files, which triggers the multipart upload within `put_object_stream`:
```rust
        let mut headers = http::HeaderMap::new();
        headers.insert("x-amz-meta-version", VERSION.into());
        let bucket = bucket.with_extra_headers(headers);
        bucket.put_object_stream(&mut rdr, path).await?
```

**Expected behavior**
The multipart upload should succeed with or without metadata.

**Environment**
 - Rust version: [e.g. `1.69`]
 - lib version [e.g. `0.34.0-beta3`]

**Additional context**
I was thinking it might be useful to introduce something like `Bucket::put_object_stream_with_metadata`. If you're open to that, I can work up a PR

*UPDATE*: after looking at the code, I thought a less intrusive approach would be just to ignore the metadata headers for any multipart-related commands other than initiate. Longer term, I think it might make more sense to replace the `_with_content_type` fn's with `_with_metadata` as content-type is considered metadata in the AWS api.

jcrossley3:
Fixes #352

### DIFF
--- a/s3/src/request/request_trait.rs
+++ b/s3/src/request/request_trait.rs
@@ -161,7 +161,6 @@ pub trait Request {
             Vec::from(content)
         } else if let Command::CompleteMultipartUpload { data, .. } = &self.command() {
             let body = data.to_string();
-            println!("CompleteMultipartUpload: {}", body);
             body.as_bytes().to_vec()
         } else if let Command::CreateBucket { config } = &self.command() {
             if let Some(payload) = config.location_constraint_payload() {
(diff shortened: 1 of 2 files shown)


durch:
@jcrossley3 Thank you!

PR #357, opened 2023-08-04
Fix put_object_stream sync for small files

### LINKED ISSUE #342: `put_object_stream` sync implementation always fails for small files
**Describe the bug**

This bug is still present in the current master branch in [s3/src/bucket.rs:1193](https://github.com/durch/rust-s3/blob/7fdb685d71385152198f906068f15faaabd28592/s3/src/bucket.rs#L1193) and in v0.33 for the sync implementation.

At this line the function should always return a Result and not continue in case of success. Currently this means we will try to read more from the reader, and return an error because it is exhausted. This makes the function always fail for small objects.

**To Reproduce**

This is a systematic error with `put_object_stream` sync implementation. A unit test should exist to cover this

**Expected behavior**

`put_object_stream` should always return the actual result of its upload 

**Environment**
 - Rust version: `1.69`
 - lib version `0.33`

jsurkont:
Closes #342

### DIFF
--- a/s3/src/bucket.rs
+++ b/s3/src/bucket.rs
@@ -1190,7 +1190,7 @@ impl Bucket {
                     // Files is not big enough for multipart upload, going with regular put_object
                     self.abort_upload(&path, upload_id)?;
 
-                    self.put_object(s3_path, chunk.as_slice())?;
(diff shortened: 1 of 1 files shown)


durch:
@jsurkont thank you!

Issue #339, opened 2023-04-24
S3 Presigned Post url


joelawm:
While Trying to implement the example in [Post](https://docs.rs/rust-s3/latest/s3/bucket/struct.Bucket.html#method.presign_post) I get a `internal error: entered unreachable code` in request\request_trait.rs:200:18

**To Reproduce**
Try and use the example

**Expected behavior**
I'm not sure if it s a oversite or not officially supported.

**Environment**
 - Rust version: 1.68.2
 - lib version 0.33.0

cs-clarence:
any updates on this? I've also encountered this

joelawm:
Yea @cs-clarence so to get around this I just implemented a lambda function in AWS Node till #358 gets resolved and/or in the official aws-sdk-rust [Issue 863](https://github.com/awslabs/aws-sdk-rust/issues/863) gets added. There really isn't a way around this unfortunately!

PR #358, opened 2023-08-07
Rework presign_post to be correct

urkle:
- the presigned post needs to provide as a response the post URL and all the fields that need to be added
- redo the presign_post function to accept only a PostPolicy which is a builder of the post policy
- add a time mock to allow for better testing

@durch please provide commends on code-style choices and contracts.

Right now this is a fully working model to get presigned posts actually working based on the documentation from these pages

https://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectPOST.html
https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-HTTPPOSTConstructPolicy.html
https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-post-example.html

I even got the same exact signature when faking the dates and base64 payload as the post-example  😄  I had to fake the payload as their example has things pretty formatted and the hash is in a slightly different order.

Reference: #339

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,7 +78,7 @@ All runtimes support either `native-tls` or `rustls-tls`, there are features for
 
 |          |                                                                                                     |
 |----------|-----------------------------------------------------------------------------------------------------|
-| `POST`   | [presign_put](https://docs.rs/rust-s3/latest/s3/bucket/struct.Bucket.html#method.presign_post)      |
+| `POST`   | [presign_post](https://docs.rs/rust-s3/latest/s3/bucket/struct.Bucket.html#method.presign_post)     |
 | `PUT`    | [presign_put](https://docs.rs/rust-s3/latest/s3/bucket/struct.Bucket.html#method.presign_put)       |
 | `GET`    | [presign_get](https://docs.rs/rust-s3/latest/s3/bucket/struct.Bucket.html#method.presign_get)       |
 | `DELETE` | [presign_delete](https://docs.rs/rust-s3/latest/s3/bucket/struct.Bucket.html#method.presign_delete) |
@@ -141,4 +141,3 @@ Each `GET` method has a `PUT` companion `sync` and `async` methods are generic o
 [dependencies]
 rust-s3 = "0.33"
 ```
-
--- a/s3/Cargo.toml
+++ b/s3/Cargo.toml
@@ -68,6 +68,7 @@ maybe-async = { version = "0.2" }
 md5 = "0.7"
 percent-encoding = "2"
 serde = "1"
+serde_json = "1"
 serde_derive = "1"
 quick-xml = { version = "0.28", features = ["serialize"] }
 sha2 = "0.10"
--- a/s3/src/bucket.rs
+++ b/s3/src/bucket.rs
@@ -38,12 +38,14 @@ use tokio::io::AsyncRead;
 use futures::io::AsyncRead;
 
 use crate::error::S3Error;
+use crate::post_policy::PresignedPost;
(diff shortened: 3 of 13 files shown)


urkle:
@durch Do you have any feedback on this?

One change I'm thinking about is making the field maps a Cow<str> instead to reduce copies for largely static strings.

urkle:
I pushed a fixup commit that switches to Cow<str> for the PostPolicy.  let me know if that is better and I'll squash the commits down.  

FYI.  I'm using this all locally in a project. It seems all of the S3 libraries for rust seem to lack PresignedPost support even amazon's.

yallxe:
Any updates on this? What is the status of this PR?

urkle:
@yallxe you can point your Cargo.toml to this branch to test it locally and verify it works for you as well (vs just me).

Add this section to your Cargo.toml
```toml
[patch.crates-io]
rust-s3 = { git = "https://github.com/urkle/rust-s3", branch = "feat-fix-presigned-post" }

```

Leonils:
Hello @urkle. Nice work on this api.
 
One detail though, I think the documentation test does not compile. Am I wrong? Following your unit test structure, I would rather assume that

* the policies values are expected to be Cows smart pointers, not strings
* the result is a structure, not just a tuple, and the fields do not implement the `Display` trait

```diff
    /// Get a presigned url for posting an object to a given path
    ///
    /// # Example:
    ///
    /// ```no_run
    /// use s3::bucket::Bucket;
    /// use s3::creds::Credentials;
    /// use s3::post_policy::*;
    /// use http::HeaderMap;
    /// use http::header::HeaderName;
+    /// use std::borrow::Cow;
    ///
    /// let bucket_name = "rust-s3-test";
    /// let region = "us-east-1".parse().unwrap();
    /// let credentials = Credentials::default().unwrap();
    /// let bucket = Bucket::new(bucket_name, region, credentials).unwrap();
    ///
    /// let post_policy = PostPolicy::new(86400).condition(
    ///     PostPolicyField::Key,
-    ///     PostPolicyValue::StartsWith("user/user1/".to_string())
+    ///     PostPolicyValue::StartsWith(Cow::Borrowed("user/user1/"))
    /// );
    ///
-    /// let (url, fields) = bucket.presign_post(post_policy).unwrap();
+   /// let PresignedPost { url, fields, .. } = bucket.presign_post(post_policy).unwrap();
-    /// println!("Presigned url: {}, fields: {}", url, fields);
+    /// println!("Presigned url: {}, fields: {?:}", url, fields);
    /// ```
```

Leonils:
Just for further context, in the AWS API, 

- PUT object is the default api to create new objects, passing the metadata in HTTP headers or query parameters.
- POST is an alternative to PUT when you want to use raw HTML forms, so metadata are part of the body. In that way they are quite different from other presigned routes (get, put, delete), as they do not have the signature in the url alone.

urkle:
> One detail though, I think the documentation test does not compile. Am I wrong? Following your unit test structure, I would rather assume that
> 
>     * the policies values are expected to be Cows smart pointers, not strings
> 
>     * the result is a structure, not just a tuple, and the fields do not implement the `Display` trait

Thanks for catching this.  Originally I had just strings and switched it to Cow in a followup commit but forgot to re-run example tests.  And at one point it was a tuple, but I reworked it as things became more complex.

I'll update these doc examples and ensure ALL tests are passing :)

durch:
@urkle wow, great work, thanks for your patience, I come back to this repo in bursts, this looks great

urkle:
🕺

PR #360, opened 2023-08-17
Move creation of the HTTP client from response() to the Bucket

### LINKED ISSUE #315: implementation of PutBucketCors
Add function to add CORS rules to bucket - like this implementation in rusoto: https://rusoto.github.io/rusoto/rusoto_s3/trait.S3.html#tymethod.put_bucket_cors

### LINKED ISSUE #347: Connection pooling support
Currently, every request results in a new TCP connection to the origin, which is very expensive due to the multiple round trips required to establish a connection to the origin.

Normally, a crate like this would store the hyper client used by a given bucket within the bucket object, allowing for a connection pool to be established by hyper to the origin. I assume that surf works similarly. For origins which support h2, this would lead to only one outbound TCP connection being needed for communicating with the origin, keeping latencies minimal.

Depending on the details of the operation and the properties of the network path between the client and the server, this may lead to substantial reductions in operation latency.

This is currently one my major blockers for using this crate, as requests I make via this crate are very slow due to the unnecessary round trips.

lyokha:
This must dramatically improve performance in high-loaded environments.
This must fix #347 (at least, for the tokio backend).

Note that
- this is implemented only for the *tokio backend*,
- TLS config related code was removed as it has been effectively disabled in commit f2a95e0, *Implement PutBucketCors, fix tokio-backend, closes #315*, and I don't know if this was made intentionally or by mistake.

### DIFF
--- a/s3/src/bucket.rs
+++ b/s3/src/bucket.rs
@@ -47,6 +47,11 @@ use crate::utils::{error_from_response_data, PutStreamResponse};
 use http::header::HeaderName;
 use http::HeaderMap;
 
+#[cfg(feature = "with-tokio")]
+use hyper::{Client, client::HttpConnector};
+#[cfg(feature = "with-tokio")]
+use hyper_tls::HttpsConnector;
+
 pub const CHUNK_SIZE: usize = 8_388_608; // 8 Mebibytes, min is 5 (5_242_880);
 
 const DEFAULT_REQUEST_TIMEOUT: Option<Duration> = Some(Duration::from_secs(60));
@@ -89,6 +94,8 @@ pub struct Bucket {
     pub extra_headers: HeaderMap,
     pub extra_query: Query,
     pub request_timeout: Option<Duration>,
+    #[cfg(feature = "with-tokio")]
+    pub(crate) client: Client<HttpsConnector<HttpConnector>>,
     path_style: bool,
     listobjects_v2: bool,
 }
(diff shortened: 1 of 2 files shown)


durch:
@lyokha, not opposed to the change in principle, but in practice, I'd prefer something like bring your own client, or similar, gonna give it some thought

durch:
Implemented in https://github.com/durch/rust-s3/commit/7d7c57b5cc967299ebf338337c7fb3b68e57c03c

lyokha:
Tested this (with *axum* as web server), looks good! Except it throws sometimes errors due to #337:

```
2023-10-16T10:41:18.466317Z  INFO request: Reading path tmp/hello.txt method=GET uri=/read/tmp/hello.txt version=HTTP/1.1
2023-10-16T10:41:18.466373Z  INFO request: Reading path tmp/hello.txt method=GET uri=/read/tmp/hello.txt version=HTTP/1.1
2023-10-16T10:41:18.466404Z  INFO request: Reading path tmp/hello.txt method=GET uri=/read/tmp/hello.txt version=HTTP/1.1
2023-10-16T10:41:18.466430Z ERROR request: S3 connection error: Could not get Read lock on Credentials method=GET uri=/read/tmp/hello.txt version=HTTP/1.1
2023-10-16T10:41:18.466458Z ERROR request: S3 connection error: Could not get Write lock on Credentials method=GET uri=/read/tmp/hello.txt version=HTTP/1.1
2023-10-16T10:41:18.466507Z  INFO request: finished processing request latency=0 ms status=502 method=GET uri=/read/tmp/hello.txt version=HTTP/1.1
```

In my case, the fix from #343 works, so it would be nice to include this into the new release. This would let using it reliably in high-loaded envs.

PR #363, opened 2023-10-10
remove check for is_ec2 when finding credentials (fixes #362)

### LINKED ISSUE #362: Enable getting credentials from http://169.254.169.254:80 when not on ec2

I would like for `aws-creds` to pick up the credentials from `http://169.254.169.254:80` if present, even if not running on ec2.  We use a system that sets up AWS credentials on this port on developer machines, which works fine with e.g. `aws s3 cp`, because it looks at `http://169.254.169.254:80` and finds the credentials as if our developer machines were on ec2.  In order for `rust-s3` to be useful to us, we'd like to be able to pick up those credentials, so I can use it to build tools that are widely and easily used within my organization.

I'd like for the `is_ec2()` call to be removed (or made in some way optional) from `from_instance_metadata()`.

**Describe alternatives you've considered**

There could be a separate constructor that omits the check for `is_ec2()`.  This could be something like Credentials::from_hologram()` potentially, although it would also work on ec2.

**Additional context**

The system we're using is called hologram: https://github.com/AdRoll/hologram

droundy:
I just removed the `is_ec2` and changed the first error to return `NotEc2`, so we would return a nice error in cases where the endpoint is not present at all.

### DIFF
--- a/aws-creds/src/credentials.rs
+++ b/aws-creds/src/credentials.rs
@@ -339,14 +339,10 @@ impl Credentials {
                         .json()?
                 }
                 Err(_) => {
-                    if !is_ec2() {
-                        return Err(CredentialsError::NotEc2);
-                    }
-
                     let role = attohttpc::get(
                         "http://169.254.169.254/latest/meta-data/iam/security-credentials",
                     )
-                    .send()?
+                    .send().map_err(|_| CredentialsError::NotEc2)?
                     .text()?;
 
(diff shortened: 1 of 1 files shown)


durch:
@droundy not sure about the change, the `send` can likely fail for a number of reasons, and always mapping it to EC2 seems worse then what is currently there

durch:
Closing for now

PR #365, opened 2023-10-23
Fix Bytes version requirement

Ten0:
0.34.0-rc4 doesn't compile with a version lower than this due to the absence of `impl From<Bytes> for Vec<u8>`.

The way to ensure this does not happen anymore is to have a CI run with [cargo-minimal-versions](https://crates.io/crates/cargo-minimal-versions).


durch (approved):

durch:
@Ten0 thank you as usual :)

PR #371, opened 2023-12-06
Fix fail-on-err description

randomairborne:
According to [github search](https://github.com/search?q=repo%3Adurch%2Frust-s3%20fail-on-err&type=code), fail-on-error does not actually add any panic! code-paths, just ones which return errors.


durch (approved):

durch:
@randomairborne nice one, thank you!

Issue #369, opened 2023-11-18
Compiling with "tokio-rustls-tls" and default-features = false still depends on native-tls


lucasmerlin:
**Describe the bug**
After updating to rust-s3 "0.34.0-rc4" from "0.33" rust-s3 now includes native-tls as a dependency even though I added the "tokio-rustls-tls" feature and disabled the default features. 

**To Reproduce**
Add the following to a empty rust project:
rust-s3 = { version = "0.34.0-rc4", default-features = false, features = ["tokio-rustls-tls"] }

run cargo tree and you should see native-tls included somewhere:

<details><summary>Details</summary>
<p>

```
rust-playground v0.1.0 (/Users/lucasprivat/IdeaProjects/rust-playground)
└── rust-s3 v0.34.0-rc4
    ├── async-trait v0.1.74 (proc-macro)
    │   ├── proc-macro2 v1.0.69
    │   │   └── unicode-ident v1.0.12
    │   ├── quote v1.0.33
    │   │   └── proc-macro2 v1.0.69 (*)
    │   └── syn v2.0.39
    │       ├── proc-macro2 v1.0.69 (*)
    │       ├── quote v1.0.33 (*)
    │       └── unicode-ident v1.0.12
    ├── aws-creds v0.36.0
    │   ├── attohttpc v0.26.1
    │   │   ├── http v0.2.11
    │   │   │   ├── bytes v1.5.0
    │   │   │   ├── fnv v1.0.7
    │   │   │   └── itoa v1.0.9
    │   │   ├── log v0.4.20
    │   │   ├── rustls v0.21.9
    │   │   │   ├── log v0.4.20
    │   │   │   ├── ring v0.17.5
    │   │   │   │   ├── getrandom v0.2.11
    │   │   │   │   │   ├── cfg-if v1.0.0
    │   │   │   │   │   └── libc v0.2.150
    │   │   │   │   └── untrusted v0.9.0
    │   │   │   │   [build-dependencies]
    │   │   │   │   └── cc v1.0.83
    │   │   │   │       └── libc v0.2.150
    │   │   │   ├── rustls-webpki v0.101.7
    │   │   │   │   ├── ring v0.17.5 (*)
    │   │   │   │   └── untrusted v0.9.0
    │   │   │   └── sct v0.7.1
    │   │   │       ├── ring v0.17.5 (*)
    │   │   │       └── untrusted v0.9.0
    │   │   ├── serde v1.0.192
[...]

PR #372, opened 2023-12-17
fix: rustls causing openssl to be built

TroyKomodo:
This commit fixes rustls feature being dependant on native-tls and binding to libcrypto.

fixes: #369

### DIFF
--- a/s3/Cargo.toml
+++ b/s3/Cargo.toml
@@ -43,10 +43,8 @@ path = "../examples/gcs-tokio.rs"
 async-std = { version = "1", optional = true }
 async-trait = "0.1"
 attohttpc = { version = "0.26", optional = true, default-features = false }
-aws-creds = { version = "0.36", default-features = false }
-# aws-creds = { path = "../aws-creds", default-features = false }
-aws-region = "0.25.4"
-# aws-region = {path = "../aws-region"}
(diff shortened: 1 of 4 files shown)


TroyKomodo:
working on this feature made me realize the structure of this crate is not rust 2021 and also uses some strange tactics like a compile feature to toggle ssl verify. Perhaps we could create a feature branch and get a 1.0 version fixing these mistakes. I do not mind putting in the work to fix and refactor them.

TroyKomodo:
I am particularly referring to using feature flags to change interface. How rust feature flags work is if any dependency in the compile tree enables the flag the feature is enabled for all users of the dep. Meaning if I use a dep which uses rust-s3 sync and I use rust-s3 with Tokio it won't work. Fortunately this has not happened but ideally we should have a better mechanism for handling this.

durch:
@TroyKomodo very nice, thank you! 

I hear you on the design choices, this is a pretty old codebase and it shows, if you'd like to come up with a set of changes you'd like to see for 1.0, I'd be happy to look them through, figure out timelines...

mdecimus:
Hi @durch 

This issue with `openssl` being compiled when using the `rustls` backend is breaking cross compilation upstream. Do you have plans to release a `0.34` point release including this fix or should we wait for `0.35`?

Thanks!

PR #380, opened 2024-03-06
Make Read and Write generics friendlier to dynamic dispatch

r-ml:
By relaxing size constraint.
Avoid double indirection. Instead of having to pass `&mut &mut dyn AsyncWrite`, we can pass `&mut dyn AsyncWrite`, for example. The use case for dynamic dispatch is object-safe traits. Static dispatch is unaffected.

`tokio` and `std` do the same thing, for example [tokio::io::copy](https://docs.rs/tokio/latest/tokio/io/fn.copy.html) and [std::io::copy](https://doc.rust-lang.org/stable/std/io/fn.copy.html).

[Example of the problem and desired outcome.](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=b329ee2e751f2af663dc2fd72c362fdb)

### DIFF
--- a/s3/src/bucket.rs
+++ b/s3/src/bucket.rs
@@ -908,13 +908,17 @@ impl Bucket {
     /// # }
     /// ```
     #[maybe_async::async_impl]
-    pub async fn get_object_range_to_writer<T: AsyncWrite + Send + Unpin, S: AsRef<str>>(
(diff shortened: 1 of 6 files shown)


durch:
@r-ml nice one! Thank you!

PR #384, opened 2024-04-19
Make BucketOwner `DisplayName` optional

erdoganege:
According to: https://docs.aws.amazon.com/AmazonS3/latest/API/API_Owner.html#API_Owner_Contents, only certain regions support `DisplayName` for the bucket. This causes problem for my bucket (eu-north-1) once I use `bucket.exists()`, it throws the following Error:
`S3(SerdeXml(Custom("missing field DisplayName")))`
Thus, it should be `Option<String>` to support all regions.


durch:
@erdoganege good, I've fixed tests up and its good to go

PR #389, opened 2024-05-28
Support custom queries when generating presigned put URLs

pepicrft:
When generating pre-signed URLs for uploading multipart parts, the query must include the parameters `partNumber` and `uploadId`. Unfortunately, this is not possible with the current interface of this Crate, which assumes that signed URLs don't need to pass custom queries.

This PR adjusts the interface to accept custom queries, too.

### DIFF
--- a/s3/src/bucket.rs
+++ b/s3/src/bucket.rs
@@ -259,6 +259,7 @@ impl Bucket {
         path: S,
         expiry_secs: u32,
         custom_headers: Option<HeaderMap>,
+        custom_queries: Option<HashMap<String, String>>
     ) -> Result<String, S3Error> {
         validate_expiry(expiry_secs)?;
         let request = RequestImpl::new(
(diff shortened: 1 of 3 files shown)


s3/src/bucket.rs
```diff
@@ -259,6 +259,7 @@ impl Bucket {
         path: S,
         expiry_secs: u32,
         custom_headers: Option<HeaderMap>,
+        custom_queries: Option<HashMap<String, String>>
```
pepicrft, line 262: How does one usually go about making this backward-compatible in Rust? I thought I'd be able to use a default value here, but it's not possible in Rust.

durch: @pepicrft you'd pass in `None` at the call site, I'll clean it up, thanks for the contribution :)

durch:
@pepicrft thank you!

PR #397, opened 2024-08-08
Add bytes_mut and into_bytes for ResponseData

kristianhasselknippe:
I needed these functions to avoid unnecessary copy so hopefully adding them is ok.


durch:
@kristianhasselknippe thank you

PR #406, opened 2024-10-29
Add a way to load Credentials from a profile file at custom location.

mrobakowski:
It is now possible to load Credentials from locations other than `$HOME/.aws/credentials`.

aws-cli allows you to customize where it looks for the `credentials` file using the `AWS_SHARED_CREDENTIALS_FILE` environment variable ([docs](https://docs.aws.amazon.com/sdkref/latest/guide/file-location.html#file-location-change)). We're using that at my $JOB. The way I implemented it here doesn't look at that environment variable, just takes an arbitrary path. That is enough for my specific use-case, but it might be reasonable to also look at the aforementioned variable in `from_profile`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/durch/rust-s3/406)

### DIFF
--- a/aws-creds/Cargo.toml
+++ b/aws-creds/Cargo.toml
@@ -37,3 +37,4 @@ rustls-tls = ["http-credentials", "attohttpc/tls-rustls"]
 [dev-dependencies]
 env_logger = "0.11"
 serde_json = "1"
+tempfile = "3"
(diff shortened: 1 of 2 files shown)


durch:
@mrobakowski Thank you for the patience, merged, and will be released in 0.37.*

Issue #404, opened 2024-10-14
Make `put_object_stream` for large files progressively upload several chunks at a time


kskalski:
**Is your feature request related to a problem? Please describe.**
I noticed that the implementation of `_put_object_stream_with_content_type` doing multi-part upload reads up all the chunks first (creating upload futures) and only then awaits for them to execute.
This blows up memory usage when uploading large files. 

Additionally I *suspect* it might be the cause of frequent "Broken pipe 32" errors that I'm seeing (could it be that we initiate request to s3, but before we actually start sending data, that is after reading all chunks from disk to memory, the server closes connection out of impatience?).

**Describe the solution you'd like**
It should incrementally add chunks (as their content is read to memory) for multi-part upload while awaiting those already created.
This kind of queue scheduling is not the most trival with rust async, but doable e.g. with https://docs.rs/futures/latest/futures/stream/struct.FuturesUnordered.html (or possibly with multi-receiver channels like `flume`) 
 
**Describe alternatives you've considered**
I suppose for now the way to go is to switch to `initiate_multipart_upload` and `put_multipart_chunk` on the user side, re-implementing `_put_object_stream_with_content_type`

**Additional context**
related or not, the main issue I'm fighting right now are `Broken pipe` and error like that:
```
reqwest::Error { kind: Request, url: "https://...amazonaws.com/...", source: hyper_util::client::legacy::Error(SendRequest, hyper::Error(IncompleteMessage)) }
```
will report back if I find something more about it, but I think it's consistent with the hypothesis that we initiate request, but not send body fast enough.

kskalski:
Confirmed that putting parts one by one using multi-part upload API for large files fixes the `broken pipe` and `incomplete message` errors for me.

So the simplest workaround is:
```rust
let multi = bucket.initiate_multipart_upload(dst, CONTENT_TYPE).await?;
let mut parts = vec![];
loop {
    let chunk = s3::utils::read_chunk_async(&mut file).await?;
    if chunk.is_empty() {
        break;
    }
    let nr = parts.len() as u32 + 1;
    tracing::debug!(len = chunk.len(), multi.upload_id, nr, "Putting part");
    let part = bucket
        .put_multipart_chunk(
            chunk,
            &multi.key,
            nr,
            &multi.upload_id,
            CONTENT_TYPE,
        )
        .await?;
    parts.push(part);
}
tracing::debug!(multi.upload_id, num_parts = parts.len(), "Completing");
let completed = bucket
[...]

PR #407, opened 2024-11-06
fix: perform streaming upload one chunk at a time

kskalski:
Closes https://github.com/durch/rust-s3/issues/404

### DIFF
--- a/s3/src/bucket.rs
+++ b/s3/src/bucket.rs
@@ -1566,11 +1566,10 @@ impl Bucket {
             chunk = next_chunk;
 
             // Analyze the response to fast fail bad upload
-            let response_data = response?;
-            if !(200..300).contains(&response_data.status_code()) {
+            if !(200..300).contains(&response.status_code()) {
                 // if chunk upload failed - abort the upload
                 self.abort_upload(&path, upload_id).await?;
-                return Err(error_from_response_data(response_data)?);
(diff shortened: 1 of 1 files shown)


durch:
@kskalski appreciate your patience, will be release in 0.37.*, very nice work, thank you!

durch:
Thank you for identifying this important memory issue and providing a fix! 

You're absolutely right that the current implementation reads ALL chunks into memory before starting uploads, which causes memory exhaustion for large files and connection timeouts.

Your sequential approach correctly fixes the memory issue. However, we're concerned about the performance impact of processing chunks one at a time, especially for large files or high-latency connections.

## Better Solution We'll Implement: Bounded Parallelism

We'll implement a sliding window approach with bounded parallelism that:
- Maintains a small number of chunks in flight simultaneously (e.g., 3-5)
- As each chunk upload completes, reads and starts the next one
- Keeps memory usage bounded while preserving upload performance

This can be done using `futures::stream::FuturesUnordered` with a buffer limit or a semaphore-based approach.

Thank you again for identifying this issue and demonstrating the problem clearly. We'll build on your work to create a solution that addresses both the memory and performance concerns.

durch:
After some back and forth, I think  I'll go with a different solution, since the sequential processing could result in a significant perf hit

PR #410, opened 2024-12-26
feat: add impl AsyncRead for ResponseDataStream

TaQuanMinhLong:
Hi, I think `ResponseDataStream` is missing the implementation of `AsyncRead ` trait, which allows us to pass the stream to a reader instead of putting the whole file in memory

- - -
This change is [Reviewable](https://reviewable.io/reviews/durch/rust-s3/410)


durch:
@TaQuanMinhLong good call out, but I'll implement it a bit differently, there are some issues here, and a lack of tests, but good callout

Issue #411, opened 2025-01-12
Issues connecting to LocalStack


JonathanWoollett-Light:
**Describe the bug**

I'm running LocalStack for testing. To test this I am opening a connection and creating an S3 bucket.

I have some Python code which does this correctly:
```python
import boto3

# Configure the S3 client to use LocalStack
s3_client = boto3.client(
    's3',
    endpoint_url='http://localhost:4566',
    aws_access_key_id='test',
    aws_secret_access_key='test',
    region_name='us-east-1'
)

# Create a bucket
bucket_name = "test-bucket"
try:
    s3_client.create_bucket(Bucket=bucket_name)
    print(f"Bucket '{bucket_name}' created successfully.")
except Exception as e:
    print(f"An error occurred: {str(e)}")
```
while my Rust code breaks:
```rust
#[tokio::main]
async fn main() {
    let region = s3::Region::Custom {
        region: String::from("us-east-1"),
        endpoint: String::from("http://127.0.0.1:4566"),
    };

    let credentials = awscreds::Credentials {
        access_key: Some(String::from("test")),
        secret_key: Some(String::from("test")),
        security_token: None,
        session_token: None,
        expiration: None,
    };
    let bucket = "test-bucket";
    let _create_bucket_response = s3::Bucket::create_with_path_style(
        bucket,
        region.clone(),
        credentials.clone(),
        s3::BucketConfiguration::public(),
    )
    .await
    .unwrap();
    let _handle = s3::Bucket::new(bucket, region, credentials).unwrap().with_path_style();
    
}
```
with:
```
thread 'main' panicked at src/main.rs:23:6:
called `Result::unwrap()` on an `Err` value: HttpFailWithBody(400, "<?xml version='1.0' encoding='utf-8'?>\n<Error><Code>InvalidLocationConstraint</Code><Message>The specified location-constraint is not valid</Message><RequestId>dce21a3a-661c-43b6-a11f-1ce14a185ce0</RequestId><LocationConstraint>us-east-1</LocationConstraint></Error>")
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
error: process didn't exit successfully: `target\debug\testing.exe` (exit code: 101)
```

**To Reproduce**
Steps to reproduce the behavior (ideally with a minimal code sample):

1. [Install LocalStack](https://github.com/localstack/localstack?tab=readme-ov-file#install).
2. Run `localstack start`
[...]

PR #412, opened 2025-01-12
Remove region restriction

JonathanWoollett-Light:
See https://github.com/durch/rust-s3/issues/411


durch:
@JonathanWoollett-Light this seems like a problem with LocalStack, I'll address it a bit differently

PR #431, opened 2025-08-08
Feat: add list_parts function

glegendr:
add list_parts function as described here
https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListParts.html

- - -
This change is [Reviewable](https://reviewable.io/reviews/durch/rust-s3/431)


durch:
@glegendr thank you for the contribution, I wonder if you could add a test, along the lines of those for methods, maybe reuse one of the multipart upload test, or just add an assertion in there

PR #434, opened 2025-10-21
Ranges are closed, so the check should be start <= end, not start < end

uckelman-sf:
The check in the ranged read functions is correct for a half-open range, but off-by-one for the closed ranges that the read functions are expecting. This prevents single-byte reads from working, because in a single-byte read, `start == end`. (Yes, single byte reads from S3 are stupid, but they should still work.)

This PR fixes the range checks.

- - -
This change is [Reviewable](https://reviewable.io/reviews/durch/rust-s3/434)

## Summary by CodeRabbit

* **Bug Fixes**
  * Adjusted S3 object range request validation to accept requests where start and end values are equal, enabling zero-length range retrieval.

### DIFF
--- a/s3/src/bucket.rs
+++ b/s3/src/bucket.rs
@@ -1186,7 +1186,7 @@ impl Bucket {
         end: Option<u64>,
     ) -> Result<ResponseData, S3Error> {
         if let Some(end) = end {
-            assert!(start < end);
+            assert!(start <= end);
         }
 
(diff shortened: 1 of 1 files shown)


uckelman-sf:
The AI-generated code review is wrong. The change does _not_ permit zero-length requests---`end` is inclusive, so when `start == end` the length is 1.

durch:
@uckelman-sf thank you, published in 0.37.1

PR #442, opened 2025-11-26
Make the HTTP client backend used by Bucket pluggable

vandry:
Traits are defined which backends must implement:

- The HTTP client `tower::Service` interface.
- The `ResponseBody` that the backend must return.

Minimal new dependencies are taken, notably `http-body` but not the whole `tower` crate, only the small crate with the trait definition.

All 3 built-in backends, Reqwest, Surf, and Atto, implement the necessary traits. Exactly one of them is compiled depending on feature flags and exposed as `DefaultBackend` so that everything works the same as before by default.

Future work will be 2 refactor some functionality such as retries and the `fail-on-err` feature as `tower::Layer` middleware.

This is a somewhat breaking change and so will require a semver bump but I tried to keep it as unintrusive as possible.

- - -
This change is [Reviewable](https://reviewable.io/reviews/durch/rust-s3/442)

## Release Notes

* **Breaking Changes**
  * Updated Bucket type and related public method signatures; code using Bucket will require updates.
  * Added new error type variant for invalid HTTP status codes.

* **New Features**
  * Added new builder methods for fluent configuration of bucket settings.
  * Enhanced error handling with improved status code validation.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/s3/src/bucket.rs
+++ b/s3/src/bucket.rs
@@ -46,7 +46,7 @@ use crate::creds::Credentials;
 use crate::region::Region;
 #[cfg(any(feature = "with-tokio", feature = "with-async-std"))]
 use crate::request::ResponseDataStream;
-use crate::request::backend::DefaultBackend;
+use crate::request::backend::{Backend, DefaultBackend};
 use crate::request::{
     ResponseBody, ResponseData, build_presigned, build_request, response_data,
     response_data_to_writer,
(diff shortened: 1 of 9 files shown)


durch:
@vandry thanks for the contribution and the considerable effort that went into this refactor. I've got a few of concerns over these changes:

1. Breaking change - Making Bucket generic (`Bucket<B>`) is a significant API break that affects all
downstream users, requiring them to update their code even if they don't need pluggable backends.
2. Complexity vs benefit tradeoff - This adds substantial complexity (generic parameters propagating
throughout the codebase) and new dependencies (`tower-service`, `http-body`, `http-body-util`) for a use
case that most users won't need.
3. Maintenance burden - The added abstraction layer makes the codebase harder to maintain and reason
about.

I wonder whats the use case here, since it seems that by utilising the features that are available you should be able to achieve a similar benefit

PR #446, opened 2026-01-03
add custom_headers parameter to initiate_multipart_upload

ZzIsGod1019:
This PR adds support for custom HTTP headers in the `initiate_multipart_upload` method, allowing users to pass custom headers (such as cache-control, metadata, storage class, etc.) when initiating multipart uploads to S3.

This feature aligns the `initiate_multipart_upload` API with the existing `put_object_with_headers` functionality, providing a consistent interface for setting custom headers across different upload methods.

- - -
This change is [Reviewable](https://reviewable.io/reviews/durch/rust-s3/446)

## Summary by CodeRabbit

* **New Features**
  * Custom HTTP headers can now be specified when initiating multipart uploads.

* **Breaking Changes**
  * Multipart upload initiation method signatures updated to include a custom headers parameter.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/s3/src/bucket.rs
+++ b/s3/src/bucket.rs
@@ -1664,7 +1664,7 @@ impl Bucket {
         }
 
         let msg = self
-            .initiate_multipart_upload(s3_path, content_type)
+            .initiate_multipart_upload(s3_path, content_type, custom_headers)
             .await?;
         let path = msg.key;
         let upload_id = &msg.upload_id;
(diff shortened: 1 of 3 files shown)


durch:
Thanks for the contribution. I’m going to close this PR rather than keep it open as backlog.

The use case — custom headers when initiating multipart uploads — is reasonable, but this PR is not in a mergeable shape for `rust-s3`:

- it changes the public `initiate_multipart_upload` signature, which is a breaking API change for existing callers
- it does not include tests for the new multipart-header behavior
- it does not add documentation/examples for the new public API shape
- it includes an unrelated signing/header change for `PutBucketLifecycle`, which should be separated from the multipart-header feature

If you want to revisit this, a better direction would be a fresh PR that keeps backward compatibility, for example by adding a separate `initiate_multipart_upload_with_headers` method or a builder-style API, with tests and docs. I’d be happy to look at a focused version along those lines.

PR #452, opened 2026-04-06
Fix ranged GET signing failure with Cloudflare R2

w4nderlust:
## Problem

`get_object_range` (ranged GET) requests fail with `SignatureDoesNotMatch` on Cloudflare R2.

The `headers()` method in `request_trait.rs` had a skip-list of commands that should not include `content-length` and `content-type` headers. `GetObjectRange` was missing from this list, so it fell through to the catch-all `_ =>` arm that adds `content-length: 0` and `content-type: text/plain`.

These empty/meaningless headers get included in the AWS4-HMAC-SHA256 canonical request. AWS S3 happens to tolerate this, but Cloudflare R2 rejects the signature because the canonical request doesn't match what R2 computes server-side.

The same class of bug also affected other body-less commands routed through PUT/POST/DELETE (`DeleteObject`, `AbortMultipartUpload`, `InitiateMultipartUpload`, etc.).

## Fix

Added a `has_body()` helper on `Command` that returns `true` only for commands that actually serialize request content (`PutObject`, `UploadPart`, `CompleteMultipartUpload`, `CreateBucket`, `PutBucketLifecycle`, `PutBucketCors`, `PutObjectTagging`). The header insertion now checks `has_body()` instead of using a skip-list or matching on the HTTP verb. This prevents the same bug from happening if new body-less commands are added in the future.

## Testing

All existing tests pass. Verified the fix against a real Cloudflare R2 bucket: ranged GET downloads that were failing with `SignatureDoesNotMatch` now succeed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/durch/rust-s3/452)

## Summary by CodeRabbit

* **Refactor**
  * Request header logic updated so copy-source is handled separately and content headers are only added for requests that carry a body.

* **Behavior**
  * Broadened recognition of which S3 operations include a request body (uploads, multipart flows, bucket creation and certain bucket settings), reducing signature mismatches for ranged/other body-less requests.

* **Tests**
  * Added tests validating body-detection and signing/header behavior.

### DIFF
--- a/s3/src/command.rs
+++ b/s3/src/command.rs
@@ -244,6 +244,7 @@ impl<'a> Command<'a> {
                 | Command::CreateBucket { .. }
                 | Command::PutBucketLifecycle { .. }
                 | Command::PutBucketCors { .. }
+                | Command::DeleteObjects { .. }
         )
     }
 
@@ -460,5 +461,27 @@ mod tests {
             upload_id: "u",
         };
         assert!(upload.has_body());
+    }
+
+    /// `DeleteObjects` is a `POST` with an XML body listing the keys to
+    /// delete. It must be reported as body-bearing so `Content-Length`
+    /// reflects the payload size and `Content-Type: application/xml` is
+    /// signed; otherwise providers reject the request or the signature.
+    #[test]
+    fn delete_objects_has_body() {
+        use crate::serde_types::{DeleteObjectsRequest, ObjectIdentifier};
+        let cmd = Command::DeleteObjects {
+            data: DeleteObjectsRequest {
+                objects: vec![ObjectIdentifier {
+                    key: "a".to_string(),
+                    version_id: None,
+                }],
+                quiet: false,
+            },
+        };
+        assert!(cmd.has_body());
+        assert_eq!(cmd.http_verb(), HttpMethod::Post);
+        assert!(cmd.content_length().unwrap() > 0);
+        assert_eq!(cmd.content_type(), "application/xml");
     }
 }
--- a/s3/src/command.rs
+++ b/s3/src/command.rs
@@ -233,6 +233,22 @@ impl<'a> Command<'a> {
     /// stray `Content-Length: 0` / `Content-Type: text/plain` headers do
     /// not enter the AWS4-HMAC-SHA256 canonical request, which Cloudflare
     /// R2 rejects as a signature mismatch (notably for ranged `GET`s).
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use s3::command::Command;
+    ///
+    /// assert!(Command::PutObject {
+    ///     content: b"hi",
+    ///     content_type: "text/plain",
+    ///     custom_headers: None,
+    ///     multipart: None,
+    /// }
+    /// .has_body());
+    /// assert!(!Command::GetObject.has_body());
+    /// assert!(!Command::GetObjectRange { start: 0, end: Some(1023) }.has_body());
+    /// ```
     pub fn has_body(&self) -> bool {
         matches!(
             self,
--- a/s3/src/command.rs
+++ b/s3/src/command.rs
@@ -215,14 +215,31 @@ impl<'a> Command<'a> {
         }
     }
 
-    /// Whether this command carries a request body that should be reflected
-    /// in `Content-Length` and `Content-Type` headers during signing.
+    /// Whether this command should include `Content-Length` and `Content-Type`
+    /// headers in the signed request.
+    ///
+    /// Returns `true` for commands that serialize a request body, plus
+    /// `InitiateMultipartUpload`. The latter is a `POST` with an empty body
+    /// but is included because:
+    ///
+    /// - Google Cloud Storage rejects the request with HTTP 411 if
+    ///   `Content-Length` is omitted from a `POST`, even when the body is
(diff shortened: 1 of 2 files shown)


durch:
Thanks for the R2 signing fix — I verified locally that this branch fixes the targeted R2 provider failure:

```bash
cd s3
cargo test streaming_big_r2_put_head_get_delete_object -- --ignored --nocapture
```

On this PR (`3ae1dd0`), that test passes.

However, I also found a deterministic GCS regression in the same local environment. The targeted GCS streaming multipart test passes repeatedly on current `master`, but fails repeatedly on this PR:

```bash
cd s3
cargo test streaming_big_gc_put_head_get_delete_object -- --ignored --nocapture
```

Observed results:

- `master` / `08545c5`: passed 3/3 runs
- this PR / `3ae1dd0`: failed 4/4 runs

The failure is:

```text
HttpFailWithBody(411, "... POST requests require a Content-length header. ...")
```

The likely cause is the new `Command::has_body()` gate. It omits `Content-Length` / `Content-Type` unless the command is listed as body-bearing:

```rust
Command::PutObject { .. }
    | Command::PutObjectTagging { .. }
    | Command::UploadPart { .. }
    | Command::CompleteMultipartUpload { .. }
    | Command::CreateBucket { .. }
    | Command::PutBucketLifecycle { .. }
    | Command::PutBucketCors { .. }
```

But `InitiateMultipartUpload` is not included. The failing GCS path uses streaming multipart upload, which sends a `POST` initiate-multipart request. GCS rejects that request when `Content-Length` is omitted, even if the body is empty.

So I agree with the direction — avoiding empty signed body headers fixes R2 — but this PR needs one more adjustment before merge:

- keep omitting empty `Content-Length` / `Content-Type` for truly body-less requests that break R2 signatures
- preserve required `Content-Length` behavior for provider-sensitive POST flows such as `InitiateMultipartUpload`
- ideally add provider-agnostic regression coverage for both cases:
  - ranged GET / body-less commands do not get bogus signed body headers
  - initiate multipart upload still gets the headers required by GCS-compatible behavior

Once that is fixed, this should be a strong compatibility improvement.

ghost:
### Brownian Motion (Brass)
  
  **Recommendation:** Proceed
  
  **Summary:** Fixes signature mismatch for ranged GET requests on Cloudflare R2.
  
  **Highlights**
- Good test coverage
- Clear commit history

**Unknowns**
- No linked issue
- Potential edge cases with other providers

**Next actions**
- Keep: the has_body() helper for future-proofing against new commands.
- Drop: any redundant checks or complexity in header management.
- Add: documentation on the rationale for has_body() and its implications.

**Reflection questions**

w4nderlust:
Thanks for the careful message. You're right, the `has_body()` gate was too narrow and dropped `Content-Length` from the `InitiateMultipartUpload` POST, which GCS rejects with HTTP 411.

I've pushed a fix and rebased the branch onto current `master`:

**Change** (`s3/src/command.rs`): added `Command::InitiateMultipartUpload` to `has_body()`. The helper's doc comment now spells out both reasons it governs Content-Length / Content-Type:
- the command actually serializes a body, **or**
- the command is a body-less `POST` whose provider-side contract still requires `Content-Length` (currently just `InitiateMultipartUpload` for GCS).

I left the rest of the body-less `POST`/`PUT`/`DELETE` commands out of `has_body()` on purpose, the original skip-list behaviour was that `DeleteObject`, `AbortMultipartUpload`, etc. did get `Content-Length: 0` via the `_ =>` arm, but you didn't observe a regression on those, and keeping them omitted preserves the R2 signing fix for the broadest surface. Happy to widen the helper if you'd rather match the previous behaviour exactly for DELETEs.

Also worth noting that `InitiateMultipartUpload`'s `Content-Type` isn't a description of the (empty) request body, it's the content type to associate with the eventual multipart object. So signing it for that command is semantically correct, not just GCS-driven.

**Provider-agnostic regression coverage** added in `s3/src/command.rs` under `mod tests`:

- `ranged_get_does_not_have_body` — pins the R2 fix (ranged GET / GET / HEAD / ListBuckets all report `false`).
- `delete_and_copy_do_not_have_body` — pins body-less DELETE / CopyObject behaviour.
- `initiate_multipart_upload_has_body_for_gcs_compat` — pins the GCS fix (`has_body() == true`, `http_verb() == POST`, `content_length() == 0`).
- `body_bearing_commands_have_body` — `PutObject` / `UploadPart` still report a body.

Verification on my end:

```
cargo test --lib   # 63 passed, 0 failed, 25 ignored
```

I don't have a GCS bucket set up locally to re-run `streaming_big_gc_put_head_get_delete_object` directly, so if you can re-run that against the rebased branch (`52881b0`) along with `streaming_big_r2_put_head_get_delete_object`, that should confirm both providers are happy.

ghost:
### Brownian Motion (Brass)
  
  **Recommendation:** Proceed
  
  **Summary:** Fixes signature mismatch for ranged GET requests on Cloudflare R2.
  
  **Highlights**
- Good test coverage
- Clear commit history

**Unknowns**
- No linked issue
- Unclear impact on other providers

**Next actions**
- Keep: the `has_body()` helper for future-proofing against new commands.
- Drop: unnecessary complexity in header management.
- Add: documentation on the implications of this change for future developers.

**Reflection questions**

PR #453, opened 2026-04-06
Fix presign tests: use fake credentials instead of minio env vars

w4nderlust:
## Problem

`make ci` (and `cargo test`) fails on a clean checkout because 4 presign tests panic trying to read `MINIO_ACCESS_KEY_ID` from the environment:

- `test_presign_get`
- `test_presign_put`
- `test_presign_post`
- `test_presign_delete`

## Fix

These tests only exercise local signing logic and never hit the network, so they don't actually need a running minio instance. Switched them from `test_minio_bucket()` (which reads env vars) to a new `test_presign_bucket()` helper with hardcoded fake credentials, following the same pattern already used by `test_presign_url_standard_ports`.

This keeps the tests in the default (non-ignored) suite so they provide CI coverage for presign PUT/POST/GET/DELETE.

## Verification

`make ci` passes clean after this change.

- - -
This change is [Reviewable](https://reviewable.io/reviews/durch/rust-s3/453)

## Summary by CodeRabbit

* **Tests**
  * Presign-related tests updated to use a new local test bucket helper targeting http://localhost:9000; no test assertions or public APIs changed.

Note: Internal testing update only; no user-facing changes.

### DIFF
--- a/s3/src/bucket.rs
+++ b/s3/src/bucket.rs
@@ -3142,6 +3142,28 @@ mod test {
         .with_path_style()
     }
 
+    /// Bucket with hardcoded fake credentials for tests that only exercise
+    /// local signing logic and never hit the network.
+    fn test_presign_bucket() -> Box<Bucket> {
+        Bucket::new(
+            "rust-s3",
+            Region::Custom {
+                region: "us-east-1".to_owned(),
+                endpoint: "http://localhost:9000".to_owned(),
+            },
+            Credentials::new(
+                Some("test_access_key"),
+                Some("test_secret_key"),
+                None,
+                None,
+                None,
+            )
+            .unwrap(),
+        )
+        .unwrap()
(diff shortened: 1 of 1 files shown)


durch:
Thanks for fixing these presign tests. This removes the accidental dependency on Minio environment credentials and makes the presign test coverage deterministic with fake credentials, which is exactly what these unit-style signing tests need.

Local verification before merge:

- `make ci` passed on `fix/presign-tests-missing-ignore` at `181664d`
- `make ci-all` reached ignored/provider tests; remaining failures were explicit provider cordons unrelated to this test-only PR:
  - local Minio unavailable at `localhost:9000`
  - existing R2 signing failures on the pre-#452 baseline

This also unblocks the current GitHub CI baseline for follow-up PRs that are failing on missing presign test credentials. Merging now.

PR #454, opened 2026-04-13
fix: Add new feature flags to aws-creds and rust-s3 to allow for ring support

dVeon-loch:
Currently, enabling rustls TLS in aws-creds (`rustls-tls`) and rust-s3
(`tokio-rustls-tls`) hardcodes the aws-lc-rs crypto provider via
`attohttpc/tls-rustls`. This makes it impossible for downstream crates
to use ring as their sole crypto provider, both providers end up
compiled into the binary.

This PR adds ring-specific feature flags:

- `aws-creds`: `rustls-tls-ring` - uses `attohttpc/tls-rustls-webpki-roots-ring` instead of `attohttpc/tls-rustls`
- `rust-s3`: `tokio-rustls-tls-ring` - uses `aws-creds/rustls-tls-ring` instead of `aws-creds/rustls-tls`

The existing `rustls-tls` and `tokio-rustls-tls` features are unchanged
for backwards compatibility.

## Why this matters

Compiling two crypto providers increases build times and binary size
with no runtime benefit. Crates that already use ring (e.g. via reqwest
0.12 or an explicit `rustls/ring` dep) currently have no way to avoid
pulling in aws-lc-sys transitively through rust-s3.

- - -
This change is [Reviewable](https://reviewable.io/reviews/durch/rust-s3/454)

## Summary by CodeRabbit

* **New Features**
  * Added `rustls-tls-ring` feature flag for alternative TLS configuration in the `aws-creds` crate
  * Added `tokio-rustls-tls-ring` feature flag for async TLS support in the `s3` crate

### DIFF
--- a/aws-creds/Cargo.toml
+++ b/aws-creds/Cargo.toml
@@ -33,6 +33,7 @@ http-credentials = ["attohttpc"]
 native-tls = ["http-credentials", "attohttpc/tls"]
 native-tls-vendored = ["http-credentials", "attohttpc/tls-vendored"]
 rustls-tls = ["http-credentials", "attohttpc/tls-rustls"]
+rustls-tls-ring = ["http-credentials", "attohttpc/tls-rustls-webpki-roots-ring"]
 
 [dev-dependencies]
 env_logger = "0.11"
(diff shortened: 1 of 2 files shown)


durch:
Thanks for the PR — the use case makes sense, and avoiding duplicate crypto providers is a reasonable goal for downstream users.

Before I can merge this, I need a bit more verification because TLS/crypto feature flags are part of the supported compatibility matrix here:

1. Please confirm the new `attohttpc/tls-rustls-webpki-roots-ring` feature is available for the `attohttpc` version currently pinned by this workspace.
2. Please add the exact commands you ran for the new feature combination, ideally including:
   - `cargo check -p aws-creds --no-default-features --features rustls-tls-ring`
   - a `s3` check/test command using `tokio-rustls-tls-ring`
3. Please confirm whether this feature should be documented in the README / crate feature list.

I’m interested in the change, but I don’t want to merge a new TLS/crypto feature path without explicit feature-matrix verification.

PR #455, opened 2026-05-04
Fix HeadObject body headers

### LINKED ISSUE #450: `Bucket::head_object` executes a request with unnecessary Content-Type headers, causing problems with Caddy reverse proxy
**Describe the bug**
When doing `head` requests through `Bucket::head_object`, despite the request being a HEAD request and thus not having a body, the library sets the Content-Type and Content-Length headers, and includes them in the `SignedHeaders` part of the authorization header. It seems that the Caddy web server takes issue with having those headers on a bodyless request, and removes these or somehow messes them up, causing the request to fail.

**To Reproduce**
Minimal example that fails:
```rs
use s3::{Bucket, creds::Credentials};

const MY_KEY: &str = "<access key>";
const MY_SECRET: &str = "<secret for access key>";
// A text file containing "Hello, world!". Here manually uploaded, in real code same issue happened with
// objects uploaded through the api on the same bucket object
const MY_FILE: &str = "testfile.txt";
const ENDPOINTS: &[&str; 2] = &[
    "http://127.0.0.1:9000",
    "https://rustfs.myproject.local"
];

#[tokio::main]
async fn main() {
    for endpoint in ENDPOINTS {
        println!(" === Endpoint: {} === ", endpoint);

        let bucket = Bucket::new(
            "my-bucket",
            s3::Region::Custom { 
                region: "rustfs".to_string(), 
                endpoint: endpoint.to_string(),
            },
            Credentials::new(
                Some(MY_KEY),
                Some(MY_SECRET),
                None, None, None
            ).unwrap(),
        ).unwrap().with_path_style();
    
        let object = bucket.get_object(MY_FILE).await.unwrap();
        println!("Get result: {}", object.to_string().unwrap());
        // For both: "Get result: Hello, world!"
    
        let head_result = bucket.head_object(MY_FILE).await;
        match head_result {
            Ok((_res, size)) => {
                // Runs for the direct-ip connection
                println!("Ok, size = {}", size);
            }
            Err(e) => {
                // Runs for the through-proxy connection
                println!("Err, msg = {}", e.to_string());
                // Err, msg = Err(HttpFailWithBody(403, ""))
            }
        }
    }
}
```

If I change the code in `index.crates-io-<blabla>\rust-s3-0.37.1\src\request\request-traits.rs`, in `pub trait Request`, in `async fn headers`, to include a special case for the HeadObject command:
```rs
// ... other special cases ...
            Command::GetBucketLocation => {}
            Command::ListBuckets => {}
            Command::HeadObject => {} // This is added
            _ => {
                headers.insert(
                    CONTENT_LENGTH,
                    self.command().content_length()?.to_string().parse()?,
                );
                headers.insert(CONTENT_TYPE, self.command().content_type().parse()?);
            }
// ... code ...
```
the example code succeeds on both calls and returns expected values.

**Expected behavior**
I am not sure. I am a big S3 noob, and have very little knowledge on how S3 APIs should behave. This could be a bug on the side of Caddy, since it should not be removing headers from request. But, this is a sensible removal - the HEAD request should have no content, so I get why Caddy does that. I have only tested the bug with a local RustFS instance, but I doubt other S3-compatible providers care about these headers.

**Environment**
 - Rust version: rustc 1.95.0-nightly (a33907a7a 2026-02-14)
 - lib version: 0.37.1
 - Caddy version: 2.10.2
 - RustFS version: 1.0.0-alpha.85

**Additional context**
If necessary I can test it out on other reverse proxies, but I do not have them setup at the moment.

durch:
## Linked issue
Fixes #450.

## Problem
`Bucket::head_object` builds its request through the common request header path. The generic fallback branch added `Content-Length: 0` and `Content-Type: text/plain` for commands that are not explicitly exempted. `GetObject` was exempted, but `HeadObject` was not, so HEAD requests emitted and signed unnecessary body headers.

## Solution
Add `Command::HeadObject` to the same narrow no-body-header branch used by `GetObject`/listing-style commands. This avoids PR #452's broader `has_body()` style change and leaves upload, multipart initiation/completion, GCS-related multipart paths, and other body-bearing commands untouched.

## Focused verification
- `cargo test -p rust-s3 --no-default-features --features sync head_object_omits_body_headers_from_request_and_signature -- --nocapture`
  - Result: passed; 1 passed, 0 failed.
- `cargo test -p rust-s3 --no-default-features --features sync request::blocking::tests -- --nocapture`
  - Result: failed in two pre-existing path-style host assertions; the new `head_object_omits_body_headers_from_request_and_signature` test passed. The failures assert path-style host behavior and are unrelated to HeadObject/body headers.

## Risk notes
- Signing risk: intentionally changes the HeadObject canonical signed header set by removing `content-length` and `content-type`; regression coverage asserts those headers are absent both from the request and from `SignedHeaders` while preserving `x-amz-content-sha256`.
- Provider/runtime cordons: change is in the shared request trait header construction, so it applies across sync/tokio/async-std backends. No URL construction, credential resolution, security token handling, request payload SHA256, multipart upload, or GCS-specific multipart behavior was changed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/durch/rust-s3/455)

## Summary by CodeRabbit

* **Bug Fixes**
  * HEAD Object requests now correctly exclude Content-Length and Content-Type from both the sent request and the signature, improving API compliance.

* **Tests**
  * Added a unit test that verifies HEAD Object requests omit body-related headers from request headers and the signed headers list.

### DIFF
--- a/s3/src/request/blocking.rs
+++ b/s3/src/request/blocking.rs
@@ -240,4 +240,29 @@ mod tests {
 
         Ok(())
     }
+
+    #[test]
+    fn head_object_omits_body_headers_from_request_and_signature() -> Result<()> {
+        let region = "custom-region".parse()?;
+        let bucket = Bucket::new("my-first-bucket", region, fake_credentials())?;
+        let request = AttoRequest::new(&bucket, "/my-first/path", Command::HeadObject).unwrap();
+
+        let headers = request.headers().unwrap();
+
+        assert!(!headers.contains_key("Content-Length"));
+        assert!(!headers.contains_key("Content-Type"));
+        assert!(headers.contains_key("x-amz-content-sha256"));
+
+        let authorization = headers.get("Authorization").unwrap().to_str()?;
+        let signed_headers = authorization
+            .split("SignedHeaders=")
+            .nth(1)
+            .and_then(|value| value.split(',').next())
+            .unwrap();
+
+        assert!(!signed_headers.contains("content-length"));
+        assert!(!signed_headers.contains("content-type"));
+
+        Ok(())
+    }
 }
--- a/s3/src/request/request_trait.rs
+++ b/s3/src/request/request_trait.rs
@@ -711,6 +711,7 @@ pub trait Request {
             }
             Command::ListObjects { .. } => {}
             Command::ListObjectsV2 { .. } => {}
+            Command::HeadObject => {}
             Command::GetObject => {}
             Command::GetObjectTagging => {}
             Command::GetBucketLocation => {}


ghost:
### Brownian Motion (Brass)
  
  **Recommendation:** Proceed
  
  **Summary:** PR effectively addresses unnecessary headers in HeadObject requests.
  
  **Highlights**
- Good test coverage
- Addresses a clear bug
- Maintains existing functionality

**Next actions**
- Keep: Tests for header omission
- Drop: Unused headers in other commands
- Add: Documentation on header behavior for HeadObject

**Reflection questions**

ghost:
### Brownian Motion (Brass)
  
  **Recommendation:** Proceed
  
  **Summary:** PR effectively addresses unnecessary headers in HEAD requests.
  **Risk:** Medium   ·   **Confidence:** 85%
  
  **Highlights**
- Good test coverage
- Addresses a clear bug
- Maintains existing functionality

**Next actions**
- Keep: The focused test for HEAD requests.
- Drop: Any unrelated changes or comments.
- Add: Documentation on the impact of this change.

**Reflection questions**

durch:
Thanks for the focused fix. This keeps the header/signing change narrow: `HeadObject` now follows the existing no-body request behavior without adopting the broader `has_body()` approach that caused provider compatibility problems elsewhere.

Verification before merge:

- local `make ci` passed in the issue worktree
- refreshed GitHub checks are green after #453 restored the presign-test baseline
- reviewed under triage rules: PASS, with the main risk being provider-sensitive signing/header behavior; the change is intentionally limited to `HeadObject`

Merging this as the fix for #450.

PR #457, opened 2026-05-04
Fix ResponseDataStream small-buffer reads

### LINKED ISSUE #437: `ResponseDataStream`: `ReaderStream` vs `.bytes`
**Describe the bug**
It looks like the `AsyncRead` implementation on `ResponseDataStream` is incorrect. The only datapoint I have at the moment is that I consistently only get about 1/3 the number of bytes back from reading all of `tokio_util::io::ReaderStream::new(response_data_stream)` vs `response_data_stream.bytes`.

Converting a `impl Stream<Item<Result<bytes::Bytes, _>>>` to a `impl AsyncRead` and then reading it should yield all the same bytes as just reading the stream directly, and I'm guessing it's not an issue with the `ReaderStream` implementation. I'll do some more testing.

**To Reproduce**
```rust
let req = bucket.get_object_stream(key).await.unwrap();
let stream = req.bytes;

let mut data = Vec::new();
while let Some(Ok(chunk)) = stream.next().await {
    data.extend(&chunk[..]);
}
println!("data.len() = {}", data.len());

// same as before, but now with:
let stream = tokio_util::io::ReaderStream::new(req);
```

**Expected behavior**
These two options produce the same number of bytes

**Environment**
 - Rust version: `rustc 1.93.0-nightly (f37aa9955 2025-10-26)`
 - lib version `0.37.0`

durch:
Linked issue: Fixes #437

## Problem
`ResponseDataStream` implemented Tokio `AsyncRead` and async-std `Read` by copying only the bytes that fit in the caller-provided buffer from each stream chunk. If a stream chunk was larger than the read buffer, the unread tail of that chunk was dropped, causing data loss for small-buffer consumers.

## Solution
When a chunk is larger than the caller buffer, preserve the unread tail by re-inserting it at the front of the boxed byte stream before continuing with the original stream. This keeps the existing `ResponseDataStream::bytes()` API and public struct fields intact while fixing both Tokio and async-std read implementations.

## Focused verification
- `cargo test -p rust-s3 --no-default-features --features tokio-native-tls request::request_trait::tests::test_async_read -- --nocapture` — passed: 4 passed, 0 failed.
- `cargo test -p rust-s3 --no-default-features --features async-std-native-tls request::request_trait::async_std_tests::test_async_read -- --nocapture` — passed: 4 passed, 0 failed.

## Risk notes
Low-to-moderate risk area because this touches streaming read semantics. The change is intentionally scoped to `ResponseDataStream` read adapters and does not alter request construction, signing, credentials, provider selection, TLS, or `ResponseDataStream::bytes()`.

## Provider/runtime cordons
- Provider behavior: no provider-specific behavior changed.
- Runtime behavior: focused coverage was run for Tokio/native-tls and async-std/native-tls read implementations.

- - -
This change is [Reviewable](https://reviewable.io/reviews/durch/rust-s3/457)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where streamed response data could be partially discarded when reads used buffers smaller than incoming chunks; sequential reads now preserve and return the full payload.

* **Tests**
  * Updated tests to validate that reading in loops returns the complete response for different async runtimes.

### DIFF
--- a/s3/src/request/request_trait.rs
+++ b/s3/src/request/request_trait.rs
@@ -21,11 +21,10 @@ use http::header::{
 };
 use std::fmt::Write as _;
 
-#[cfg(feature = "with-async-std")]
-use async_std::stream::Stream;
-
-#[cfg(feature = "with-tokio")]
-use tokio_stream::Stream;
+#[cfg(any(feature = "with-tokio", feature = "with-async-std"))]
+use futures_util::Stream;
+#[cfg(any(feature = "with-tokio", feature = "with-async-std"))]
+use futures_util::stream::{self, StreamExt};
 
 #[derive(Debug)]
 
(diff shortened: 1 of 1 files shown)


ghost:
### Brownian Motion (Brass)
  
  **Recommendation:** Refactor
  
  **Summary:** PR addresses data loss in ResponseDataStream; approach is sound but complex.
  
  **Highlights**
- Good test coverage
- Addresses a clear bug

**Unknowns**
- Impact on performance with large streams

**Next actions**
- Keep: Test cases that validate the fix
- Drop: Unused imports and redundant comments
- Add: Documentation for the new behavior

**Reflection questions**

ghost:
### Brownian Motion (Brass)
  
  **Recommendation:** Refactor
  
  **Summary:** PR addresses data loss in small-buffer reads, but complexity may outweigh benefits.
  **Risk:** Medium   ·   **Confidence:** 70%
  
  **Highlights**
- Good test coverage
- Clear commit history

**Unknowns**
- Impact on performance with larger streams
- Long-term maintainability of the new stream handling approach

**Next actions**
- Keep: tests that verify correct behavior with small buffers
- Drop: unnecessary complexity in handling stream re-insertion
- Add: documentation explaining the rationale behind changes

**Reflection questions**

durch:
Thanks for the data-integrity fix. The previous `AsyncRead` implementation could drop bytes when a caller-provided read buffer was smaller than the next stream chunk; this preserves chunk remainders and adds focused coverage for small-buffer reads.

Verification before merge:

- local `make ci` passed in the issue worktree
- refreshed GitHub checks are green after #453 restored the presign-test baseline
- reviewed under triage rules: PASS; high-impact streaming area, but the fix is localized and covered by targeted regression tests

Merging this as the fix for #437.

PR #465, opened 2026-06-02
fix: exclude content-length and content-type from DeleteObject signed…

r-oswald:
`DeleteObject` is missing from the skip-list in `request_trait.rs` `headers()`, so requests get `content-length: 0` and `content-type: text/plain` signed.

  L7 proxies strip `Content-Length: 0` per RFC 7230 §3.3.2 so you get a signature mismatch.

  So this merge request strips this header on delete request.

- - -
This change is [Reviewable](https://reviewable.io/reviews/durch/rust-s3/465)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of delete requests to ensure they are properly formatted without unnecessary headers, fixing an issue where delete operations may have been incorrectly processed.

### DIFF
--- a/s3/src/request/request_trait.rs
+++ b/s3/src/request/request_trait.rs
@@ -741,6 +741,7 @@ pub trait Request {
             Command::GetObjectTagging => {}
             Command::GetBucketLocation => {}
             Command::ListBuckets => {}
+            Command::DeleteObject => {}
             _ => {
                 headers.insert(
                     CONTENT_LENGTH,


ghost:
### Brownian Motion (Brass)
  
  **Recommendation:** Proceed
  
  **Summary:** PR addresses header signing issue for DeleteObject requests effectively.
  
  **Highlights**
- Clear identification of the problem
- Concise implementation
- Relevant RFC reference

**Unknowns**
- Potential impact on existing clients using DeleteObject

**Next actions**
- Keep: the change to exclude content-length and content-type headers
- Drop: any additional complexity not related to the core fix
- Add: tests to confirm the behavior of DeleteObject requests

**Reflection questions**

sigmaris:
I believe this would fix issue #466 (I arrived here via [this bug report](https://support.stalw.art/t/bug-report-s3-purge-blob-store-task-fails-with-403-signaturedoesnotmatch-on-cloudflare-r2/444)).

PR #467, opened 2026-07-02
fix: do not sign Content-Length/Content-Type on bodyless DELETE requests

### LINKED ISSUE #466: DeleteObject signs `content-length` and `content-type` headers, causing 403 SignatureDoesNotMatch on Cloudflare R2
**Describe the bug**
When deleting an object, the request includes content-length: 0 and content-type: text/plain in the SigV4 signature. A DELETE has no body, so these headers should not be part of the canonical request. AWS S3 tolerates this, but Cloudflare R2 enforces SigV4 more strictly and rejects the request with 403 SignatureDoesNotMatch ("The request signature we calculated does not match the signature you provided"). The result is that no object can ever be deleted on R2, while PUT, GET and HEAD all work fine.

**To Reproduce**
Configure a bucket against a Cloudflare R2 endpoint with credentials, put an object, then call bucket.delete_object(path). The PUT succeeds; the DELETE returns 403 SignatureDoesNotMatch. The same call against AWS S3 succeeds, which is why it often goes unnoticed.

**Environment**
 - lib version [e.g. `0.22.12`]: 0.37.2

eldon-databanx:
Fixes #466.

## Problem

`headers()` in `request_trait.rs` falls through to the catch-all arm for DELETE commands and inserts `Content-Length: 0` and `Content-Type: text/plain` into the signed header map. Since a DELETE has no body, the HTTP client never actually transmits those headers, so the canonical request signed by rust-s3 differs from the one the server reconstructs.

AWS S3 tolerates the discrepancy, which is why this goes unnoticed there. Cloudflare R2 enforces SigV4 strictly and rejects every DELETE with `403 SignatureDoesNotMatch` — PUT/GET/HEAD work fine, but no object can ever be deleted on R2.

## Fix

Skip inserting `Content-Length`/`Content-Type` for all bodyless DELETE commands: `DeleteObject`, `DeleteObjectTagging`, `AbortMultipartUpload`, `DeleteBucket`, `DeleteBucketCors` and `DeleteBucketLifecycle`. (`DeleteObjects` is a POST with an XML body and keeps its headers.)

## Testing

Verified in production against Cloudflare R2 (Stalwart mail server, which had been unable to purge any blob on R2): with this patch, `delete_object` succeeds — a backlog of ~2.8k objects that had been failing with `SignatureDoesNotMatch` was deleted without a single error. Behaviour against AWS S3 is unchanged, since the headers were no-ops there.

- - -
This change is [Reviewable](https://reviewable.io/reviews/durch/rust-s3/467)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved request header handling for delete-style operations so they no longer include unnecessary `Content-Length` or `Content-Type` headers.
  * This makes bodyless requests more consistent and better aligned with expected service behavior.

### DIFF
--- a/s3/src/request/request_trait.rs
+++ b/s3/src/request/request_trait.rs
@@ -741,6 +741,16 @@ pub trait Request {
             Command::GetObjectTagging => {}
             Command::GetBucketLocation => {}
             Command::ListBuckets => {}
+            // Bodyless DELETE requests must not sign Content-Length/Content-Type:
+            // the HTTP client never transmits them, so strict SigV4 validators
+            // (e.g. Cloudflare R2) reconstruct a different canonical request and
+            // reject with SignatureDoesNotMatch (see issue #466).
+            Command::DeleteObject
+            | Command::DeleteObjectTagging
+            | Command::AbortMultipartUpload { .. }
+            | Command::DeleteBucket
+            | Command::DeleteBucketCors { .. }
+            | Command::DeleteBucketLifecycle => {}
             _ => {
                 headers.insert(
                     CONTENT_LENGTH,


ghost:
### Brownian Motion (Brass)
  
  **Recommendation:** Proceed
  
  **Summary:** Fixes R2 DELETE signatures by omitting headers clients never send
  **Risk:** Medium   ·   **Confidence:** 87%
  
  **Highlights**
- Targets a real interop bug with strict SigV4 providers
- Minimal one-file change aligned with existing request signing logic
- Author validated the fix against Cloudflare R2 in production

**Unknowns**
- No automated test currently proves the canonical request mismatch is fixed
- Unclear whether every backend/client combination already omits these headers for all listed commands
- The current ci status is unstable, so broader regression risk is harder to rule out

**Next actions**
- Keep: the narrow special-casing for bodyless DELETE commands
- Keep: the explanatory comment tying the change to strict SigV4 validators
- Add: a regression test that asserts DELETE canonical headers exclude content-length/content-type
- Add: verify whether any other bodyless methods with explicit zero-length bodies need the same treatment, especially presign/delete variants
- Drop: expanding scope to unrelated request builders unless they share the same signing path

**Reflection questions**
- What exact headers does each backend actually transmit for these DELETE requests?
- Are we fixing the signing model to match the wire, or just papering over one backend's behavior?
- Can the request/header derivation be made source-of-truth instead of relying on command-specific exceptions?

cvermilion:
I believe #452 also includes a (more general) fix.

nfodigital:
Still hitting this on rust-s3 0.37.2 via Stalwart v0.16.20 (released 2026-08-30) against Cloudflare R2.

`DeleteObject` includes `content-length` and `content-type` in the SigV4 canonical request. R2 validates signed headers strictly and rejects the signature; AWS S3 is lenient and accepts it, which is why this only surfaces on R2 / Ceph RGW / SeaweedFS.

Scope in practice: `GET`/`PUT`/`HEAD` are unaffected — only `DELETE` (and ranged `GET`, which goes through the same match arm, cf. #452). For us that means Stalwart's nightly `purgeBlob` task cannot delete a single expunged blob: **1280 failed DELETEs in the last 24h**, and the bucket has been growing unbounded since May.

This PR is marked mergeable/clean and has been sitting since 2026-08-06. @durch — is anything blocking it, or is there a preferred direction between this, #465, and the minimal-signed-header approach in #470? Happy to help test against a real R2 endpoint.

